PR #5, opened 2016-03-31
Remove DockerResource and use spring-cloud-deployer-resource-docker p…

trisberg:
…roject instead

Depends on spring-cloud/spring-cloud-deployer#33


markfisher:
LGTM; merged

PR #17, opened 2016-04-27
Retrieve app count from environment variables

### LINKED ISSUE #16: Retrieve app count correctly
The deployment property `spring.cloud.deployer.count` needs to be retrieved from deployment request environment variables instead of request definition.

Currently, the `countProperty` is always set to null and thereby the replication controller has the count value as `1` always.

ilayaperumalg:
- Since deployment request definition wouldn't have the `spring.cloud.deployer.count` property set, we need to retrieve it from environment variable

This resolves #16


trisberg:
Rebased and merged via commit 0f6e602a584189e81fc2090d80cf9f071e7bbc22

PR #23, opened 2016-06-02
Allow task launch to re-launch existing task/job

trisberg:
- check for existing job, delete if complete, throw exception if launching/running

- clean up imports

Resolves #22

PR #25, opened 2016-06-08
Adding command line args to the container deployment 

### LINKED ISSUE #21: Add support for command-line arguments
As a developer, I want to be able to configure command line args as a separate set of properties to be passed to an app when it's deployed.

Similar to the Local SPI (_via: spring-cloud/spring-cloud-dataflow#597_), we would need this supported at the Kubernetes deployer SPI.

trisberg:
Fixes #21

Depends on #23


mminella:
LGTM.  Merged as 8da8a57f0876f86525b9a893eee1c1da66ec7033.

PR #26, opened 2016-06-09
Adding logging for deploy

### LINKED ISSUE #24: Log any exceptions during deployment
The calling core Data Flow server code only logs the message from the exception. We should add a catch clause in the Kubernetes deployer and log more detailed exception message including the cause and stacktrace. See https://github.com/spring-cloud/spring-cloud-dataflow-server-kubernetes/issues/81

trisberg:
Fixes #24

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
@@ -76,23 +76,28 @@ public class KubernetesAppDeployer extends AbstractKubernetesDeployer implements
 
 		logger.debug("Deploying app: {}", appId);
 
-		AppStatus status = status(appId);
-		if (!status.getState().equals(DeploymentState.unknown)) {
-			throw new IllegalStateException(String.format("App '%s' is already deployed", appId));
+		try {
+			AppStatus status = status(appId);
+			if (!status.getState().equals(DeploymentState.unknown)) {
+				throw new IllegalStateException(String.format("App '%s' is already deployed", appId));
+			}
+
+			int externalPort = 8080;
+			Map<String, String> parameters = request.getDefinition().getProperties();
+			if (parameters.containsKey(SERVER_PORT_KEY)) {
+				externalPort = Integer.valueOf(parameters.get(SERVER_PORT_KEY));
+			}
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
```diff
@@ -76,23 +76,28 @@ public String deploy(AppDeploymentRequest request) {
...
-		logger.debug("Creating repl controller: {} on {}", appId, externalPort);
-		createReplicationController(appId, request, idMap, externalPort);
+			logger.debug("Creating repl controller: {} on {}", appId, externalPort);
+			createReplicationController(appId, request, idMap, externalPort);
 
-		return appId;
+			return appId;
+		} catch (Exception e) {
+			logger.error(e.getMessage(), e);
+			throw e;
```
markfisher: shouldn't this be wrapped in a `RuntimeException`?

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
```diff
@@ -132,9 +137,9 @@ public void undeploy(String appId) {
 			selector.put(SPRING_APP_KEY, appId);
 			Boolean podDeleted = client.pods().withLabels(selector).delete();
 			logger.debug("Deleted pods for: {} {}", appId, podDeleted);
-		} catch (KubernetesClientException e) {
+		} catch (Exception e) {
 			logger.error(e.getMessage(), e);
-			throw new RuntimeException(e);
+			throw e;
```
markfisher: shouldn't this be wrapped in a `RuntimeException` here also?

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
trisberg: Actually, they would already be RuntimeExceptions, should change the catch to RuntimeException and the just re-throw it. Unless we would want to introduce a new DeploymentException in core.

markfisher:
LGTM; squashed and merged as https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/commit/00848d5ec9ce7cf4f59eced8c0b2c19196876eb4

PR #36, opened 2016-07-01
[DO NOT MERGE EVER] POC to support partitioning

trisberg:
- we craete one pod per app

- we create additional containers if count > 1

- we set INSTANCE_INDEX and set server.port to port+INSTANCE_INDEX

- status is calculated based on first container for now

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.deployer.spi.kubernetes;
 
+import static io.fabric8.kubernetes.api.model.KubernetesKind.List;
+
 import java.util.HashMap;
 import java.util.Map;
 
@@ -26,6 +28,8 @@ import org.springframework.cloud.deployer.spi.app.DeploymentState;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
 
 import io.fabric8.kubernetes.api.model.Container;
+import io.fabric8.kubernetes.api.model.DoneablePod;
+import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.PodList;
 import io.fabric8.kubernetes.api.model.PodSpec;
 import io.fabric8.kubernetes.api.model.PodSpecBuilder;
@@ -35,6 +39,8 @@ import io.fabric8.kubernetes.api.model.ResourceRequirements;
 import io.fabric8.kubernetes.api.model.Service;
 import io.fabric8.kubernetes.api.model.ServiceSpecBuilder;
 import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.dsl.ClientPodResource;
+import io.fabric8.kubernetes.client.dsl.FilterWatchListDeletable;
 
 /**
  * A deployer that targets Kubernetes.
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
Verified the changes with `kubernetes` server and could see the stream partitioning working.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
```diff
@@ -156,8 +166,9 @@ public AppStatus status(String appId) {
 	private ReplicationController createReplicationController(
 			String appId, AppDeploymentRequest request,
 			Map<String, String> idMap, int externalPort) {
-		String countProperty = request.getDeploymentProperties().get(COUNT_PROPERTY_KEY);
-		int count = (countProperty != null) ? Integer.parseInt(countProperty) : 1;
+//		String countProperty = request.getDeploymentProperties().get(COUNT_PROPERTY_KEY);
+//		int count = (countProperty != null) ? Integer.parseInt(countProperty) : 1;
+		int count = 1;
```
ilayaperumalg: We can probably remove the variable `count` and use `1` as the value for the number of `replicas` directly.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
```diff
@@ -188,14 +199,19 @@ private PodSpec createPodSpec(String appId, AppDeploymentRequest request, int po
...
-		// add memory and cpu resource limits
-		ResourceRequirements req = new ResourceRequirements();
-		req.setLimits(deduceResourceLimits(properties, request));
-		container.setResources(req);
+		for (int index = 0; index < count; index++) {
+			Container container = containerFactory.create(appId, request, port, index);
 
-		podSpec.addToContainers(container);
+			// add memory and cpu resource limits
+			ResourceRequirements req = new ResourceRequirements();
```
ilayaperumalg: I believe `ResourceRequirement` would be same for all the instances for the given app and hence the `requirement` can be computed outside the loop.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppInstanceStatus.java
```diff
@@ -45,8 +45,8 @@ public KubernetesAppInstanceStatus(String moduleId, Pod pod, KubernetesDeployerP
 		this.moduleId = moduleId;
 		this.pod = pod;
 		this.properties = properties;
-		// we assume one container per pod
-		if (pod != null && pod.getStatus().getContainerStatuses().size() == 1) {
+		// we assume one container per pod - if more than one the status is for the first one
+		if (pod != null && pod.getStatus().getContainerStatuses().size() >= 1) {
```
ilayaperumalg: I guess this is where it gets tricky. Since the entire stream deployment status is computed based on the `ContainerStatus` I think we need to get consider checking all the ContainerStatus info. Otherwise, the stream status would be considered `deployed` even if some of the requested app instances aren't deployed yet.

trisberg:
Discussed the general solution for partitioning with INSTANCE_INDEX on both Mesos and Kubernetes with Eric earlier. We might want to use the same solution and schedule multiple individual app instances for both Mesos and Kubernetes. If use what is in this POC we'll end up with all partitions being handled on a single node, which isn't exactly what we would want. That would be OK for a demo, but not a real deployment.

trisberg:
Replaced by #38

PR #38, opened 2016-07-06
Implement scaling support to take into account whether an instance in…

### LINKED ISSUE #37: Implement scaling support to take into account whether an instance identifier is required

trisberg:
…dex is required or not

- for indexed deployments deploy N number of apps with instance index of 0 to N-1, where N is the count deployment property

- for non-indexed deployments use Kubernetes scaling to launch N number of pods for the app, where N is the count deployment property

Resolves #37

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
 		<java.version>1.8</java.version>
 		<kubernetes-client.version>1.3.83</kubernetes-client.version>
 		<kubernetes-assertions.version>2.2.110</kubernetes-assertions.version>
-		<spring-cloud-deployer-spi.version>1.0.1.RELEASE</spring-cloud-deployer-spi.version>
+		<spring-cloud-deployer-spi.version>1.0.2.BUILD-SNAPSHOT</spring-cloud-deployer-spi.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -20,7 +20,7 @@
 		<java.version>1.8</java.version>
 		<kubernetes-client.version>1.3.83</kubernetes-client.version>
 		<kubernetes-assertions.version>2.2.110</kubernetes-assertions.version>
-		<spring-cloud-deployer-spi.version>1.0.1.RELEASE</spring-cloud-deployer-spi.version>
+		<spring-cloud-deployer-spi.version>1.0.2.BUILD-SNAPSHOT</spring-cloud-deployer-spi.version>
```
ilayaperumalg: This can be `1.0.3` eventually I think.

trisberg: It will be moved to `1.0.2.RELEASE` before we release

ilayaperumalg: yeah, it is `1.0.2`. The commit description for the release had it as `1.0.3.RELEASE` in `spring-cloud-deployer` which confused me.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 85893355e9f1350f6b72b5240d679eac1bd8d68b. Also, verified scale out and stream partitioning functionalities on Google Cloud with the kafka apps

Issue #46, opened 2016-08-23
Non-default external port is appended to container args


waseemh:
App external port (default is 8080) can be modified by setting 'server.port' key in the AppDefintion properties map.
However, all properties in map are later appended into container args (implementation in [DefaultContainerFactory](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/master/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java)).
This behavior is incorrect since server.port is not a known arg when executing **_Docker run --server.port_**, thus causing an error while deploying app.

I managed to work-around this issue in [DefaultContainerFactory](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/master/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java) by checking if property key is 'server.port' before appending to container args:
```java
Map<String, String> args = request.getDefinition().getProperties();
    for (Map.Entry<String, String> entry : args.entrySet()) {
        if(!entry.getKey().equals("server.port")) { //work-around
            cmdArgs.add(String.format("--%s=%s", entry.getKey(), entry.getValue()));
        }
 }
```

waseemh:
I suggest one of following solutions for this issue:
1) Move container args representation from AppDefinition properties to AppDeploymentRequest commandlineArguments.
2) Move external port representation from AppDefinition properties to AppDeploymentRequest deploymentProperties
3) Assign a namespace in AppDefinition properties for container args. i.e: only properties with 'arg.' prefix should be appended as container args.

PR #49, opened 2016-08-29
Modifying container args handling

waseemh:
Do not add AppDefinition properties as container args
Current behavior causes issues when a non-default external port is used since server.port property is added as a container argument in DefaultContainerFactory.
Docker is unable to identify this argument thus unable to start container.

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
@@ -119,14 +119,7 @@ public class DefaultContainerFactory implements ContainerFactory {
 	 * Create command arguments
 	 */
 	protected List<String> createCommandArgs(AppDeploymentRequest request) {
-		List<String> cmdArgs = new LinkedList<String>();
-		// add properties from deployment request
-		Map<String, String> args = request.getDefinition().getProperties();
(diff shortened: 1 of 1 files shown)


waseemh:
Should fix issue: #46

trisberg:
Not sure I understand the motivation for this. We need to pass the properties to the Boot app. Could you clarify why this is a problem?

waseemh:
So you assume that Docker image will include Boot app in its entry point ?
If yes, then you can close this PR.

trisberg:
For now we do assume it is a boot app, we are however discussing supporting other apps as well, both plain Java apps and non-Java apps. If you have ideas how we can improve the SPI for these type of apps we'd be interested to hear them.

trisberg:
I guess we could add a `spring.cloud.deployer.kubernetes.container.command` property where you could specify the command to use instead of what's provided in the docker image. Maybe also add a `spring.cloud.deployer.kubernetes.container.port` to set the port, that way you don't have to use the AppDefinition properties for non-Boot apps.

waseemh:
Yes, container port and command fit best as deployment properties.

`container.port` property will be used to define container external port (also to define load balancer endpoint port and so on...)
In case `server.port` is defined in `AppDefinition` properties (Spring Boot apps), then it will override the `container.port` property.
Sounds fair?

trisberg:
Created an issue to track this change - https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/50

PR #55, opened 2016-09-20
Use bare pods and use unique id for each launch

### LINKED ISSUE #42: Explore using bare pods instead of jobs for tasks

### LINKED ISSUE #56: Generate unique id for each launch of a task

trisberg:
- switch to use bare pods instead of jobs for tasks

- generate unique id for each launch of a task

- switch to use Commons Logging

Resolves #42
Resolves #56

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,6 +48,11 @@
 			<version>${kubernetes-client.version}</version>
 		</dependency>
 		<dependency>
+			<groupId>org.hashids</groupId>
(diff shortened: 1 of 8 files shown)


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as 74e6165. Also verified the task/app launch with unique IDs and task launching that uses pods instead of jobs.

PR #64, opened 2016-10-12
Allow overriding container command and ports

wkennedy:
These commits allow the user to specify app properties to allow overriding the container entry point and/or specify additional ports.

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
@@ -17,10 +17,9 @@
 package org.springframework.cloud.deployer.spi.kubernetes;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -85,6 +84,7 @@ public class DefaultContainerFactory implements ContainerFactory {
 				.withImage(image)
 				.withEnv(envVars)
 				.withArgs(createCommandArgs(request));
+
 		if (port != null) {
 			container.addNewPort()
 					.withContainerPort(port)
@@ -96,6 +96,23 @@ public class DefaultContainerFactory implements ContainerFactory {
 							createProbe(port, properties.getLivenessProbePath(), properties.getLivenessProbeTimeout(),
 									properties.getLivenessProbeDelay(), properties.getLivenessProbePeriod()));
 		}
+
+		//Add additional specified ports.  Further work is needed to add probe customization for each port.
+		List<Integer> additionalPorts = getContainerPorts(request);
+		if(!additionalPorts.isEmpty()) {
+			for (Integer containerPort : additionalPorts) {
+				container.addNewPort()
+						.withContainerPort(containerPort)
+						.endPort();
+			}
+		}
+
+		//Override the containers default entry point with one specified during the app deployment
+		List<String> containerCommands = getContainerCommands(request);
+		if(!containerCommands.isEmpty()) {
+			container.withCommand(containerCommands);
+		}
+
 		return container.build();
 	}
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
```diff
@@ -137,7 +157,53 @@ protected Probe createProbe(Integer externalPort, String endpoint, int timeout,
...
+     * @return a list of strings that represents the command and any arguments for that command
+     */
+    private List<String> getContainerCommand(AppDeploymentRequest request) {
+        List<String> containerCommandList = new ArrayList<>();
+        String containerCommand = request.getDeploymentProperties()
+                .get("spring.cloud.deployer.kubernetes.containerCommand");
+        if (containerCommand != null) {
+			logger.info("Building container command from AppDeploymentRequest: "
+					+ containerCommand);
+            String[] containerCommandSplit = containerCommand.split(",");
```
ericbottard: Why split on comma here? This should typically mimic shell behavior IMO

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
```diff
@@ -137,7 +157,53 @@ protected Probe createProbe(Integer externalPort, String endpoint, int timeout,
...
+     * @param request AppDeploymentRequest - used to gather application overridden
+     * container command
+     * @return a list of strings that represents the command and any arguments for that command
+     */
+    private List<String> getContainerCommand(AppDeploymentRequest request) {
+        List<String> containerCommandList = new ArrayList<>();
+        String containerCommand = request.getDeploymentProperties()
+                .get("spring.cloud.deployer.kubernetes.containerCommand");
+        if (containerCommand != null) {
+			logger.info("Building container command from AppDeploymentRequest: "
```
ericbottard: All these should not be logged at info level, this is too broad. Should be debug, if not trace

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
```diff
@@ -137,7 +157,53 @@ protected Probe createProbe(Integer externalPort, String endpoint, int timeout,
...
+        if (containerPorts != null) {
+            String[] containerPortSplit = containerPorts.split(",");
+            for (String containerPort : containerPortSplit) {
+                logger.info("Adding container ports from AppDeploymentRequest: "
+                        + containerPort);
+                try {
+                    Integer port = Integer.parseInt(containerPort.trim());
+                    containerPortList.add(port);
+                } catch (NumberFormatException e) {
+                    logger.info("Value is not a valid port number.");
```
ericbottard: Should not catch the error but let it bubble up

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactoryTest.java
```diff
@@ -69,6 +72,80 @@ public void create() {
...
+                "8081, 8082, invalid");
+        AppDeploymentRequest appDeploymentRequest = new AppDeploymentRequest(definition,
+                resource, props);
+
+        Container container = defaultContainerFactory.create("app-test",
+                appDeploymentRequest, null, null);
+        assertNotNull(container);
+        List<ContainerPort> containerPorts = container.getPorts();
+        assertNotNull(containerPorts);
+        assertTrue("There should be three ports set", containerPorts.size() == 2);
```
ericbottard: Given comment above, this should be changed to assert failure. But this test was copy/pasted and `"three"` was not changed here.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
wkennedy: The way Kubernetes handles container commands is in  a comma delimited format, for instance:

```
apiVersion: v1
kind: Pod
metadata:
  name: hello-world
spec:  # specification of the pod's contents
  restartPolicy: Never
  containers:
  - name: hello
    image: "ubuntu:14.04"
    command: ["/bin/echo", "hello", "world"]
```

Grabbing a comma delimited value from the spring.cloud.deployer.kubernetes.containerPorts makes it easier to pass the command and argument list to the fabric8 container model, which expects the shell command as list of strings.

ericbottard:
Kubernetes uses YAML files extensively, so this is why it appears that the command line array is using _comma separated values_. But it's really that, an array.
We need a way to pass an array of args as well, possibly taking care of quotes etc.

Given that this feature may be valuable to many deployer implementations, I'll work on a PR adding it to the core [deployer](https://github.com/spring-cloud/spring-cloud-deployer/) project, and we can then use it in this PR. Hope that makes sense

wkennedy:
Yes, that makes sense.  Thanks!

ericbottard:
Working on changes, will merge

ericbottard:
Reworked a bit and merged as 8b96e867f1cf9ec6473d9b6dfed0a3e86c4e6a8d

Thanks @wkennedy !

PR #68, opened 2016-10-26
Allows setting the imagePullPolicy for containers

tschuettel:
The image pull policy for both apps and tasks can now be set
at deployment time by setting

```
spring.cloud.deployer.kubernetes.imagePullPolicy
```

to one of Always,IfNotPresent,Never for each app/task.

The server can be set up to have a default image pull policy by setting

```
spring.cloud.deployer.kubernetes.imagePullPolicy
```

to one of the given values for the server. The default value is Always.

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -94,4 +94,25 @@ public class AbstractKubernetesDeployer {
 		return limits;
 	}
 
+	/**
+	 * Get the image pull policy for the deployment request. If it is not present use the server default.
+	 *
+	 * @param properties The server properties.
+	 * @param request The deployment request.
+	 * @return The image pull policy to use for the container in the request.
+	 */
+	ImagePullPolicy deduceImagePullPolicy(KubernetesDeployerProperties properties, AppDeploymentRequest request) {
+		String pullPolicyOverride =
+				request.getDeploymentProperties().get("spring.cloud.deployer.kubernetes.imagePullPolicy");
+
+		ImagePullPolicy pullPolicy;
+		if (pullPolicyOverride == null) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -94,4 +94,25 @@ protected AppStatus buildAppStatus(KubernetesDeployerProperties properties, Stri
...
+	 */
+	ImagePullPolicy deduceImagePullPolicy(KubernetesDeployerProperties properties, AppDeploymentRequest request) {
+		String pullPolicyOverride =
+				request.getDeploymentProperties().get("spring.cloud.deployer.kubernetes.imagePullPolicy");
+
+		ImagePullPolicy pullPolicy;
+		if (pullPolicyOverride == null) {
+			pullPolicy = properties.getImagePullPolicy();
+		} else {
+			pullPolicy = ImagePullPolicy.valueOf(pullPolicyOverride);
```
FrontierPsychiatrist: I agree. I will add the necessary changes.

Do you prefer squashed/amended commits for corrections on PRs or keeping the history?

trisberg: Just amend the PR and I'll squash it on merge. Thanks for taking care of this.

FrontierPsychiatrist:
I added the relaxed binding of the pull policy override and the missing license and author tag.

I also added a test for the `AbstractKubernetesDeployer` like in #70 but noticed that Maven will not execute it if it is named `AbstractKubernetesDeployerTest`. Should we change that in #70? Of course that will lead to merge conflicts between these two PRS.

trisberg:
LGTM, I'll move the `AbstractKubernetesDeployerTests` to the `KubernetesDeployerTests` on merge

trisberg:
Merged as commit 1cb27cb73754a0587e2d72e8b58a814cb622a405

PR #70, opened 2016-10-26
Allows separated settings for resource limits and requests for containers

tschuettel:
Adds the possiblity to set Kubernetes resource request values additional
to the resource limits. This allows for creating burstable pods, currently the deployer only supports guaranteed pods.
For more details see [0] and [1].

They can be set serverwide and per deployment request.

The Kubernetes resource limit values get a new key to be set but for
backwards compability the old names are still available (but have lower
priority).

I.e. with this commit the following is possible to allow burstable containers
in Kubernetes:

```
spring.cloud.deployer.kubernetes.requests.cpu=500m
spring.cloud.deployer.kubernetes.requests.memory=640Mi
spring.cloud.deployer.kubernetes.limits.cpu=2000m
# Same as spring.cloud.deployer.kubernetes.cpu=2000m
spring.cloud.deployer.kubernetes.limits.memory=2048Mi
# Same as spring.cloud.deployer.kubernetes.memory=2048Mi
```

[0] https://github.com/kubernetes/kubernetes/blob/release-1.3/docs/design/resource-qos.md
[1] http://kubernetes.io/docs/user-guide/compute-resources/#how-pods-with-resource-requests-are-scheduled

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
@@ -25,6 +25,40 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 @ConfigurationProperties(prefix = "spring.cloud.deployer.kubernetes")
 public class KubernetesDeployerProperties {
 
+	/**
+	 * Encapsulates resources for Kubernetes Container resource requests and limits
+	 */
+	public static class Resources {
+
+		private String cpu;
+
+		private String memory;
+
+		public Resources() {
+		}
+
+		public Resources(String cpu, String memory) {
+			this.cpu = cpu;
+			this.memory = memory;
+		}
+
+		public String getCpu() {
+			return cpu;
+		}
+
+		public void setCpu(String cpu) {
+			this.cpu = cpu;
+		}
+
+		public String getMemory() {
+			return memory;
+		}
+
+		public void setMemory(String memory) {
+			this.memory = memory;
+		}
+	}
+
 	private static String KUBERNETES_NAMESPACE =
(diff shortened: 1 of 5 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -92,15 +126,31 @@
...
 
 	/**
+	 * Memory and CPU limits (i.e. maximum needed values) to allocate for a Pod.
+	 */
+	private Resources limits = new Resources("500m", "512Mi");
+
+	/**
+	 * Memory and CPU requests (i.e. guaranteed needed values) to allocate for a Pod.
+	 */
+	private Resources requests = new Resources("500m", "512Mi");
```
trisberg: We shouldn't default this, only set this if explicitly requested. If someone set the mem limit to "128Mi" this would fail with: ```[StatusCause(field=spec.template.spec.containers[0].resources.limits, message=Invalid value: "128Mi": must be greater than or equal to memory request, reason=FieldValueInvalid, additionalProperties={})]```

trisberg: I can change this to `new Resources();` on merge unless you see a problem doing that

FrontierPsychiatrist: Yes, didn't think about that. The default values can be removed. I think the K8S API doesn't mind empty values, right? So this does not affect the `AbstractKubernetesDeployer`, we can just pass `null`.

trisberg:
Rebased and merged as commit 8836a3b566ec623749b2ae829fb8e84e32135b60 and commit 3872b4506858f0a98c377db8d4835bef698b3a37

PR #71, opened 2016-10-26
Adds support for specifying entry point style for Docker images

### LINKED ISSUE #65: Allow customization how properties are passed when launching the Docker image
This should improve compatibility with Docker images for Spring Boot apps that can use either exec or shell form entry points.

For Boot apps we should use SPRING_APPLICATION_JSON env var for app properties.

Add a config property `spring.cloud.deployer.kubernetes.entryPointStyle` with values of `EXEC`, `SHELL` and `BOOT`. `EXEC` should be the default and then we pass properties as arguments, for `SHELL` we would pass properties as env vars, and with `BOOT` we would be using SPRING_APPLICATION_JSON for app properties.

trisberg:
- three styles EXEC (default), SHELL and BOOT
- Affects how app properties are passed in.
  -- for EXEC app properties will be passed in as command line args
  -- for SHELL they will be passed in as environment variables
  -- for BOOT they will be passed in using a SPRING_APPLICATION_JSON environment variable.
- switch to using the newer Spring Boot test annotations

Resolves #65

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
@@ -23,6 +23,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -57,6 +59,7 @@ public class DefaultContainerFactory implements ContainerFactory {
 
 	@Override
 	public Container create(String appId, AppDeploymentRequest request, Integer port, Integer instanceIndex) {
+
 		String image = null;
 		try {
 			image = request.getResource().getURI().getSchemeSpecificPart();
@@ -64,6 +67,10 @@ public class DefaultContainerFactory implements ContainerFactory {
 			throw new IllegalArgumentException("Unable to get URI for " + request.getResource(), e);
 		}
 		logger.info("Using Docker image: " + image);
+
+		KubernetesDeployerProperties.DockerEntryPointStyle entryPointStyle =
+				determineEntryPointStyle(properties, request);
+		logger.info("Using Docker entry point style: " + entryPointStyle);
 
 		Map<String, String> envVarsMap = new HashMap<>();
 		for (String envVar : properties.getEnvironmentVariables()) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
```diff
@@ -76,6 +83,28 @@ public Container create(String appId, AppDeploymentRequest request, Integer port
 		//image supports it.
 		envVarsMap.putAll(getAppEnvironmentVariables(request));
 
+		if (!(entryPointStyle == KubernetesDeployerProperties.DockerEntryPointStyle.EXEC)) {
```
ericbottard: Wouldn't a `switch` here be more readable?

ericbottard:
Apart from the `switch` comment, LGTM.

BUT, is there real value in supporting the `SPRING_APPLICATION_JSON` approach? It exists _e.g._ in CF because there were issues with escaping stuff / ENV_VARS did not support dots.

Which makes me wonder: do we properly (bash?) escape command line arguments here? Do we need to, or does k8s take care of that?

trisberg:
The BOOT option would come into play if you want to build an app and want to be able to use $JAVA_OPTS to pass in JVM settings. To do that you need to use the shell entryPoint style and that means you have to pass in properties as env vars. The most flexible env vars option is as we know SPRING_APPLICATION_JSON.

ericbottard:
I'm not sure I follow:
> The BOOT option would come into play if you want to build an app and want to be able to use $JAVA_OPTS to pass in JVM settings. To do that you need to use the `SHELL` (emphasis mine) entryPoint style and that means you have to pass in properties as env vars.

=> then you're not using `BOOT`
> The most flexible env vars option is as we know SPRING_APPLICATION_JSON.

=> but they're not passed as process ENV_VARS. They do end up un the Spring `Environment`, but that's all (in addition of being the easiest way to escape special chars. But commandLine args should allow that too, either thanks to k8s or us)

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
trisberg: modified the code to use a switch and reordered stuff a bit

trisberg:
You can't use command line args for the shell style, need exec for that. But the exec style doesn't do variable substitution of $JAVA_OPTS, need exec style for that. The EXEC, SHELL, BOOT options gives us the most flexibility. As for escaping, that's a separate issue and I'm not aware of any issues so far.

trisberg:
And the BOOT option would work with either shell or exec style docker images, for using $JAVA_OPTS we would use BOOT plus a docker image that uses shell style entryPoint.

trisberg:
@ericbottard where are we with this PR? Any additional concerns?

ericbottard:
Rebased and merged as e3cd281

PR #84, opened 2016-11-10
Create shared methods in AbstractKubernetesDeployer

trisberg:
- for setting memory/cpu

- and volumes/mounts

Resolves #83

PR #86, opened 2016-11-15
Added all supported volume types. Fixes #85.

### LINKED ISSUE #85: Add support for all supported volume types
Building on #79 which covers [hostPath](http://kubernetes.io/docs/user-guide/volumes/#hostpath) volumes. All other [volume types](http://kubernetes.io/docs/user-guide/volumes/#types-of-volumes) should also be supported (or at least all volumes that have a model generated by [kubernetes-model](https://github.com/fabric8io/kubernetes-model)).

donovanmuller:
This PR adds support for all the supported [volume types](http://kubernetes.io/docs/user-guide/volumes/#types-of-volumes) available in the [kubernetes-model](https://github.com/fabric8io/kubernetes-model).

Volumes can be defined as deployer properties only. The reasoning being that the deinfition of a Volume is generally an administrative concern (especially in the case of `PersistentVolumeClaim` etc.) on a Kubernetes cluster and not something that would be defined per app.

VolumeMounts can be defined as deployer *and* deployment properties. Allowing common volume mounts to be assigned to all apps and at an individual app level (additive).
VolumeMounts defined as deployment properties override those defined as deployer properties. The [format of the volume mounts definition](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/compare/master...donovanmuller:85-update?expand=1#diff-ac742fc1f690acd37859b18a8bbf479bR204) in deployment properties is up for debate...

See [dataflow-server.yml](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/compare/master...donovanmuller:85-update?expand=1#diff-0affb8ad59ab41d93c465650a259cc7a) for examples of how the Volumes can be defined as deployer properties.

Note: This PR does not take #84 into account but can be adjusted accordingly.

> The `Volume` and `VolumeMount` properties are the bare kubernetes-model types. I.e. there is no local wrapper class (like [HostVolumeMount](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/master/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java#L33)) for these concepts. The reason is mainly for the `Volume` type, as creating wrapper classes for all the supported volume types would be tedious and add unnecessary maintenance, keeping them in sync with the kubernetes-model types.

Also some minor cleanup of unused imports and removed some unnecessary syntax due to move to Java 8.

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
@@ -228,7 +228,7 @@ public class DefaultContainerFactory implements ContainerFactory {
 						volume.length == 3 ? Boolean.valueOf(volume[2]) : Boolean.FALSE, null));
 			}
 		}
-		properties.getVolumeMounts();
+		volumeMounts.addAll(properties.getVolumeMounts());
 		return new ArrayList<>(volumeMounts);
 	}
 
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
@@ -16,27 +16,18 @@
 
 package org.springframework.cloud.deployer.spi.kubernetes;
 
-import static java.lang.String.format;
-
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-import org.springframework.cloud.deployer.spi.app.AppDeployer;
-import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
-import org.springframework.cloud.deployer.spi.util.CommandLineTokenizer;
-import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
+import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.ContainerBuilder;
@@ -45,6 +36,13 @@ import io.fabric8.kubernetes.api.model.HTTPGetActionBuilder;
 import io.fabric8.kubernetes.api.model.Probe;
 import io.fabric8.kubernetes.api.model.ProbeBuilder;
 import io.fabric8.kubernetes.api.model.VolumeMount;
+
+import org.springframework.boot.bind.YamlConfigurationFactory;
+import org.springframework.cloud.deployer.spi.app.AppDeployer;
+import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.cloud.deployer.spi.util.CommandLineTokenizer;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * Create a Kubernetes {@link Container} that will be started as part of a
@@ -66,8 +64,7 @@ public class DefaultContainerFactory implements ContainerFactory {
 
 	@Override
 	public Container create(String appId, AppDeploymentRequest request, Integer port, Integer instanceIndex) {
-
-		String image = null;
+		String image;
 		try {
 			image = request.getResource().getURI().getSchemeSpecificPart();
 		} catch (IOException e) {
@@ -130,8 +127,7 @@ public class DefaultContainerFactory implements ContainerFactory {
 		container.withName(appInstanceId)
 				.withImage(image)
 				.withEnv(envVars)
-				.withArgs(appArgs)
-				.withVolumeMounts(getVolumeMounts(request));
+				.withArgs(appArgs);
 
 		if (port != null) {
 			container.addNewPort()
@@ -185,7 +181,7 @@ public class DefaultContainerFactory implements ContainerFactory {
 	 * Create command arguments
 	 */
 	protected List<String> createCommandArgs(AppDeploymentRequest request) {
-		List<String> cmdArgs = new LinkedList<String>();
+		List<String> cmdArgs = new LinkedList<>();
 		// add properties from deployment request
 		Map<String, String> args = request.getDefinition().getProperties();
 		for (Map.Entry<String, String> entry : args.entrySet()) {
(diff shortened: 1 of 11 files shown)


src/test/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployerTests.java
```diff
@@ -37,7 +37,7 @@
  *
  * @author Moritz Schulze
  */
-public class KubernetesDeployerTests {
+public class AbstractKubernetesDeployerTests {
```
trisberg: Tests with name AbstractXxx aren't included during regular builds, renaming to RunAbstractKubernetesDeployerTests on merge

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
```diff
@@ -189,6 +197,41 @@ protected Probe createProbe(Integer externalPort, String endpoint, int timeout,
...
+		if (!StringUtils.isEmpty(volumeMountDeploymentProperty)) {
+			String[] volumePairs = volumeMountDeploymentProperty.split(",");
+			for (String volumePair : volumePairs) {
+				String[] volume = volumePair.split(":");
+				Assert.isTrue(volume.length <= 3, format("Invalid volume mount: '{}'", volumePair));
+				volumeMounts.add(new VolumeMount(volume[1], volume[0],
+						volume.length == 3 ? Boolean.valueOf(volume[2]) : Boolean.FALSE, null));
+			}
+		}
+		properties.getVolumeMounts();
```
trisberg: I don't think this statement does anything, should't this be the returned if `volumeMountDeploymentProperty` is empty?

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
```diff
@@ -253,19 +250,8 @@ private PodSpec createPodSpec(String appId, AppDeploymentRequest request, Intege
...
-			podSpec.withVolumes(properties.getHostVolumeMounts().stream()
-					.map(mv -> {
-						Volume volume = new Volume();
-						HostPathVolumeSource hostPath = new HostPathVolumeSource();
-						hostPath.setPath(mv.getHostPath());
-						volume.setHostPath(hostPath);
-						volume.setName(mv.getName());
-						return volume;
-					}).collect(Collectors.toList()));
+		if (properties.getVolumeMounts() != null) {
```
trisberg: Isn't it a bit limiting to require volume mounts to be in the deployer configuration? If I wanted to use a `hostPath` volume I would have to re-deploy the server after re-configuring it?

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployerTests.java
donovanmuller: Ah ok, good to know.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
donovanmuller: The thinking was that volumes would be something that a cluster admin would create and would not change all that often. Happy to support volumes defined as deployment properties as well.

However, not sure how we would parse the format of the value passed in the deployment property?
I.e. we essentially get a String value which we would have to parse and build the `Volume` model with (the `volumeMount` format is consistent, hence [this format](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/86/files#diff-ac742fc1f690acd37859b18a8bbf479bR204)). The different volume types differ slightly in their configuration (e.g. `hostPath` vs. [`cephfs`](https://github.com/kubernetes/kubernetes/blob/master/examples/volumes/cephfs/cephfs.yaml#L14) so not sure what format to use to support the various volume types?

donovanmuller:
As for the downside of adding all configured volumes to all apps:

I'm not aware of any issues that would break a deployment because of that but I can see why that would be a little heavy handed. We could filter volumes based on matching volume mounts?

donovanmuller:
Based on https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/86#issuecomment-261167289, I've added a filter so that only matching volume/volumeMount pairs are added.

trisberg:
One final suggestion for the formats of volumes/volumeMounts in deployment properties.

We could allow volumes to be specified during deployment using a single line YAML format like: `"[{name: 'testhostpath', hostPath: { path: '/tmp/test'}}]"` parse this with some YAML parser

Maybe we could do the same for the volumeMounts to be consistent:
`"[{testhostpath: `/test/hostPath`}, {testpvc: '/test/pvc'}, {testnfs: '/test/nfs'}]"`

Also, looks like the deployer config volumeMounts gets added to the ones defined during deployment, should it be either or in case someone wants to override the path of a mount already defined in the deployer config?

donovanmuller:
Correct, volumeMounts specified as deploy*ment* properties override those defined as deplo*yer* properties. Hence the use of a `Set`, add all volumeMounts from deployment properties and then add all deployer properties but don't add if there is already a volumeMount (i.e. the override - logic is a little inverted I guess :)) added from the deployment properties. 

Happy with the format of the volumeMounts/volume being JSON/YAML. I'll get on that now...

trisberg:
FYI, I slapped together a quick test to verify that we could indeed parse the YAML snippet:

```
	@Test
	public void parseYaml() throws Exception {
		YamlConfigurationFactory<KubernetesDeployerProperties> yamlConfigurationFactory = new YamlConfigurationFactory<>(
				KubernetesDeployerProperties.class);
		yamlConfigurationFactory.setYaml("{ volumeMounts:" + "[{name: 'testhostpath', mountPath: '/test/hostPath'}, {name: 'testpvc', mountPath: '/test/pvc'}, {name: 'testnfs', mountPath: '/test/nfs'}]" + " }");
		yamlConfigurationFactory.afterPropertiesSet();
		KubernetesDeployerProperties props = yamlConfigurationFactory.getObject();
		System.out.println(props.getVolumeMounts());
	}
```

Looks good to me ;)

trisberg:
Aa for the Set for VolumeMounts, if I change the mountPath for an existing volumeMount during deployment it would be a different VolumeMount object so they both would get added to the set, or am I missing something?

trisberg:
Thanks Donovan, I think we have it all covered by now. I'll run some final tests and then I'll merge this.

donovanmuller:
You're correct, it would result in two VolumeMounts with the same name but different `mountPath`.
I filter by name now...

I've also added support for defining volumes with deployment properties with the new YAML format.

trisberg:
LGTM now!

A few changes on merge:

- adding a label selector for the pod we want in the `testDeploymentWithMountedHostPathVolume` test, it failed without it when I already had other pods running at the same time. 

- changing the mount/mountVolumes properties YAML snippets in tests to not have a property name prefix, adding that prefix when we parse the snippet instead. This is also how the JavaDocs have it defined.

trisberg:
Squashed and merged as commit 6f5a92948c41c0f7f073d3b91c670a9d889c0fe7

PR #91, opened 2016-11-18
Add support for new deployer cpu and memory deployment properties

### LINKED ISSUE #76: Adopt to use new memory and cpu constants
Adopt to use new memory and cpu constants added in core -
https://github.com/spring-cloud/spring-cloud-deployer/issues/69

trisberg:
Resolves #76

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -27,6 +27,7 @@ import org.springframework.boot.bind.RelaxedNames;
 import org.springframework.cloud.deployer.spi.app.AppDeployer;
 import org.springframework.cloud.deployer.spi.app.AppStatus;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.cloud.deployer.spi.util.ByteSizeUtils;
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -90,7 +91,18 @@ protected AppStatus buildAppStatus(KubernetesDeployerProperties properties, Stri
 	 * @param request    The deployment properties.
 	 */
 	protected Map<String, Quantity> deduceResourceLimits(KubernetesDeployerProperties properties, AppDeploymentRequest request) {
+		String memDeployer = getCommonDeployerMemory(request);
 		String memOverride = request.getDeploymentProperties().get("spring.cloud.deployer.kubernetes.memory");
```
ericbottard: Why is the old `spring.cloud.deployer.kubernetes.memory` property still supported? I thought the point of introducing a common property was that all implementations would use it

ericbottard: gotcha

trisberg:
Resolving the conflicts was a mess - pushing a re-worked PR instead - https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/96

PR #102, opened 2017-01-17
Add support for enabling hostNetwork for pods.

### LINKED ISSUE #99: spring.cloud.deployer.kubernetes.hostNetwork
Moved from rom https://github.com/spring-cloud/spring-cloud-dataflow-server-kubernetes/issues/161:

org.springframework.cloud.deployer.spi.kubernetes.KubernetesAppDeployer
(1.0.4.RELEASE)

This class should accept this flag : spring.cloud.deployer.kubernetes.hostNetwork
and apply to
io.fabric8.kubernetes.api.model.PodSpec

This is useful when we want the pods to be hostNetwork enabled.

Thoughts ?

trisberg:
- this allows us to deploy a stream using something like:
  stream deploy ticktock --properties "app.time.server.port=8281,app.log.server.port=8282,app.*.spring.cloud.deployer.kubernetes.hostNetwork=true"

- when using "minikube" we can now access the deployed apps on http://192.168.99.100:8281/health and http://192.168.99.100:8282/health respectively

- see https://kubernetes.io/docs/api-reference/v1/definitions/#_v1_podspec

Resolves #99

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -109,7 +109,8 @@ public class AbstractKubernetesDeployer {
 			podSpec.addNewImagePullSecret(properties.getImagePullSecret());
 		}
 
-		Container container = containerFactory.create(appId, request, port, instanceIndex);
+		boolean hostNetwork = getHostNetwork(request);
+		Container container = containerFactory.create(appId, request, port, instanceIndex, hostNetwork);
 
 		// add memory and cpu resource limits
 		ResourceRequirements req = new ResourceRequirements();
@@ -125,6 +126,9 @@ public class AbstractKubernetesDeployer {
 						.anyMatch(volumeMount -> volumeMount.getName().equals(volume.getName())))
 				.collect(Collectors.toList()));
 
+		if (hostNetwork) {
+			podSpec.withHostNetwork(true);
+		}
 		podSpec.addToContainers(container);
 
 		if (neverRestart){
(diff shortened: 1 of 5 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -308,6 +312,18 @@ protected ImagePullPolicy deduceImagePullPolicy(AppDeploymentRequest request) {
...
 
+	/**
+	 * Get the hostNetwork setting for the deployment request.
+	 *
+	 * @param request The deployment request.
+	 * @return Whether host networking is requested
+	 */
+	protected boolean getHostNetwork(AppDeploymentRequest request) {
+		String hostNetwork =
+				request.getDeploymentProperties().get("spring.cloud.deployer.kubernetes.hostNetwork");
```
ilayaperumalg: Not on the scope of this PR: but, we can try to keep a constant for `spring.cloud.deployer.kubernetes` at some point for all the deployer properties prefix.

ilayaperumalg: Currently, the property `spring.cloud.deployer.kubernetes.hostNetwork` is not part of `KubernetesDeployerProperties`. Is there a reason for it? I guess having this property there would be useful to set at the startup/app level vs deployment property only.

trisberg: Hmm, yes, might be useful, let me add it there

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -308,6 +312,26 @@ protected ImagePullPolicy deduceImagePullPolicy(AppDeploymentRequest request) {
...
+	/**
+	 * Get the hostNetwork setting for the deployment request.
+	 *
+	 * @param request The deployment request.
+	 * @return Whether host networking is requested
+	 */
+	protected boolean getHostNetwork(AppDeploymentRequest request) {
+		String hostNetworkOverride =
+				request.getDeploymentProperties().get("spring.cloud.deployer.kubernetes.hostNetwork");
+		boolean hostNetwork;
```
ilayaperumalg: Might be simpler if we do this:

```
boolean hostNetwork = properties.isHostNetwork();
		if (!StringUtils.isEmpty(hostNetworkOverride)) {
			hostNetwork = Boolean.valueOf(hostNetworkOverride);
		}
```
Can add while merging if you agree.

trisberg: I'm OK with either way - we do follow the pattern I used for other properties, but I'm OK with your way as well.

ilayaperumalg: ok, I would like to go with the pattern as well. will keep yours as is.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as 2d50d61

PR #117, opened 2017-03-16
Add configuration-processor as optional dependency

### LINKED ISSUE #116: Add config properties metadata jar
This is for the Configuration properties autocompletion in IDE.

ilayaperumalg:
- For autocompletion of properties in IDE

Resolves #116

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,11 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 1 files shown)


ericbottard:
Merged. Could be argued that some descriptions in javadoc are too long / use carriage returns etc. Maybe we could move some parts of it in regular comments or `@see` tags. cc @trisberg

PR #134, opened 2017-05-23
Adding properties for 'LoadBalancer' type sevices

### LINKED ISSUE #133: Return URL property as part of app status call

trisberg:
- adding external IP and port

- adding external URL

Resolves #133


ilayaperumalg:
@trisberg It looks like these changes are already there as part of the latest docker image for `spring-cloud-dataflow-server-kubernetes`. Wonder how that could be possible. Maybe the docker image was manually pushed from the PR branch?

The PR changes look good and able to verify against `runtime apps` as well.

Merging.

ilayaperumalg:
Merged as 89350ba

trisberg:
@ilayaperumalg thanks for merging, and, yes, I pushed the PR changes as 'latest' in Docker Hub so we could work on the acceptance tests.

Issue #150, opened 2017-08-04
Try using StatefulSets for the indexed apps


trisberg:
See: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

dturanski:
Blocked by this https://github.com/fabric8io/kubernetes-client/issues/706

dturanski:
Issue resolved. The api supports it but hard to find.

dturanski:
https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/160

mbogoevici:
I think it makes sense to support StatefulSets as a distinct deployment flavour regardless whether the applications are indexed or not (e.g. by adding a flag). Indexed apps can use StatefulSets by default. This could help supporting stateful applications (e.g. Kafka Streams) as part of the pipeline.

PR #156, opened 2017-09-14
Sidecars DO NOT MERGE. Pending #150

### LINKED ISSUE #151: Add support for defining a side-car container
Support side-car container for database access to Google Cloud SQL as well as Istio Service Mesh integration.

dturanski:
This PR addresses issue #151 adding support for multiple containers in a pod.  Sidecar apps may be written in any language and are assumed to be always running. Currently sidecars must specify the Docker image.

Each sidecar may expose more than one port. By default a TCP liveness probe is created for the first port listed. It is also possible to designate a port.

Additionally Sidecars may configure volume mounts, environment variables, command, args (the difference AFAICT is the command includes the command is a list starting with the command (e.g. `/bin/bash`) along with the arguments whereas `arguments` apply to the specified container command.

The deployment status has been enhanced to deal with multiple containers. 

Resolves #151


markpollack:
Will wait for experience from the RIFF team before adding to the k8s deployer.

PR #160, opened 2017-10-09
Deploy as StatefulSet for indexed Apps

### LINKED ISSUE #150: Try using StatefulSets for the indexed apps

dturanski:
This PR resolves #150 and a couple of related items.

* Indexed applications are deployed as StatefulSets. K8s manages creates a separate pod for each replica and automatically indexes the pod's name and HOSTNAME natively. This makes it much easier to manage indexed apps, necessary for partitioning, and also to scale indexed apps.
* `KubernetesDeploymentProperties.setCreateDeployment(), isCreateDeployment()` and the associated `KubernetesAppDeployer.createReplicationController()` are deprecated. The feature is still in place for non-indexed apps. 

TLDR; It works, but required a lot of creativity. 

* Some caveats re StatefulSet support:
  - The `statefulSet:spec:podManagementPolicy` is required to be set to `Parallel` in order to pass one of the integration tests that deploys an indexed app with 3 replicas and one configured to crash. The expected behavior is that this will be reported as a partial deployment with instance statuses ['deployed',deployed','failed'].  This is not possible with the default `podManagementPolicy=OrderedReady` which starts pods in index order and will not start the next one until the previous one is ready. This test is part of the spec for the AppDeployer SPI.
  - The current fabric8 Java KubernetesClient does not support `podManagementPolicy`. In order to set this value, it is necessary to by pass the client API and execute raw REST requests via the clients, OKHttp client. This is true for both create and delete operations. These are the only ones required.
  - StatefulSets require a `Persistent Volume Claim` used to persist application state.  Since dataflow apps are stateless, we don't really care about persistence. We create a 0 Mi persistent store. 
  - The primary advantage of StatefulSets is that it allows K8s to manage the indexing of Pods. Spring Cloud Stream Binder requires the `INSTANCE_INDEX` property to be set for partitioning. K8s does not currently expose the index. It can be derived from parsing the `HOSTNAME`. This is done by launching an InitContainer that can execute a shell command to extract the index value and write it to `config/application.properties` on a shared volume, from where the boot application can get it. This works well, but the init container takes about 20-30 seconds to do this.
  - The fabric8 client also does not know about `InitContainer`, so a "legacy" annotation is used to set the init container configuration as a JSON string. 
  - In order to figure out why `testMultipleInstancesDeploymentAndPartialState` was failing, I need some more detailed logging about specific conditions used to determine the deployment state. So I refactored the code used to determine the state of apps in the Running phase accordingly. It should be easier to read as well.

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppInstanceStatus.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppInstanceStatus.java
@@ -19,6 +19,8 @@ package org.springframework.cloud.deployer.spi.kubernetes;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 import io.fabric8.kubernetes.api.model.Service;
 import io.fabric8.kubernetes.api.model.ServicePort;
@@ -36,6 +38,7 @@ import org.springframework.cloud.deployer.spi.app.DeploymentState;
  *
  * @author Florian Rosenberg
  * @author Thomas Risberg
+ * @author David Turanski
  */
 public class KubernetesAppInstanceStatus implements AppInstanceStatus {
 
@@ -44,6 +47,8 @@ public class KubernetesAppInstanceStatus implements AppInstanceStatus {
 	private Service service;
 	private KubernetesDeployerProperties properties;
 	private ContainerStatus containerStatus;
+	private RunningPhaseDeploymentStateResolver runningPhaseDeploymentStateResolver =
+		new DefaultRunningPhaseDeploymentStateResolver();
 
 	public KubernetesAppInstanceStatus(Pod pod, Service service, KubernetesDeployerProperties properties) {
 		this.pod = pod;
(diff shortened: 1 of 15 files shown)


sabbyanandan:
Hi, @trisberg @markfisher - please let us know what you think.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppInstanceStatus.java
```diff
@@ -189,4 +150,203 @@ else if (containerStatus.getRestartCount() == 0 &&
 		}
 		return result;
 	}
+
+	static interface RunningPhaseDeploymentStateResolver {
```
viniciusccarvalho: @dturanski would it make sense to move this interface to a public package? And does it make sense to have the Resolver type to be writable via a setter? Maybe not now, but couldn't we have different strategies that could be chosen and set according to environment? Just asking, seems like a candidate to allow users to change behavior

dturanski: I'll take another look at it

viniciusccarvalho:
Merged via f0d7263c88415c8b5a00abc7eb9a3b802c2984a4

PR #171, opened 2017-12-23
Updating initContainer section

### LINKED ISSUE #170: Update initContainers Section
As per 1.8, K8S no longer support the annotation based init container method. There's new API support since 1.6. 
Currently we use the annotation method and hence why we are failing on 1.8 and passing in 1.7. This is story is to use the new API on fabric8 to create the init containers for indexed applications

viniciusccarvalho:
Fixes #170 

This fix solves the issue with the initContainer for indexed apps and kubernetes 1.8.

According to the spec, as of 1.8 version beta annotations are no longer supported. This makes the init container feature supported for any version after 1.6 see : https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#support-and-compatibility


trisberg:
LGTM, tested against 1.7.8 and 1.8.4 clusters.

Issue #164, opened 2017-11-28
Change the default memory limit to 1024Mi


trisberg:
We have to override this most of the time anyway, so we might as well increase it to `1024Mi`

markpollack:
fixed in d6c9c5983c1b1d5ee76926214dc2b11dcafe921b

PR #174, opened 2018-01-10
Update default memory from 512 to 1024

sabbyanandan:
I asked about the change to a deprecated property via slack; I haven't received any suggestions, so I'm submitting it as a proposal. 

We can discuss any concerns here. 

Resolves spring-cloud/spring-cloud-deployer-kubernetes#164.


markpollack:
rebased and merged as d6c9c5983c1b1d5ee76926214dc2b11dcafe921b

PR #181, opened 2018-02-08
Set StatefulSet's storage

### LINKED ISSUE #178: Fix issue with empty storage request when creating stateful set
Currently, the logic in deployer is looking for indexed property in the request and if finds one, it will create stateful set with empty storage request which hangs and does not create PVC in Kubernetes on AWS (i.e. EBS based storage) It seems to work fine in minikube thou.

ilayaperumalg:
- Set this via AppDeploymentRequest's deployment properties or by default 1024Mi
 - Update tests

Resolves #178

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.ContainerBuilder;
+import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaimBuilder;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaimList;
@@ -73,6 +74,7 @@ import java.util.Map;
  * @author Mark Fisher
  * @author Donovan Muller
  * @author David Turanski
+ * @author Ilayaperumal Gopinathan
  */
 public class KubernetesAppDeployer extends AbstractKubernetesDeployer implements AppDeployer {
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
```diff
@@ -314,7 +314,11 @@ private Deployment createDeployment(String appId, AppDeploymentRequest request,
 	private int getCountFromRequest(AppDeploymentRequest request) {
 		String countProperty = request.getDeploymentProperties().get(COUNT_PROPERTY_KEY);
 		return (countProperty != null) ? Integer.parseInt(countProperty) : 1;
+	}
 
+	private String getMemoryFromRequest(AppDeploymentRequest request) {
```
markpollack: This is conflating two separate variables.  The MEMORY_PROPERTY_KEY is used for RAM.  See `getCommonDeployerMemory` which also has some additional logic for parsing.  In this case we want to set the size of the storage resource, which is different than RAM.

See https://kubernetes.io/docs/concepts/policy/resource-quotas/  "Storage Resource Quota" and https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#components 

```  volumeClaimTemplates:
  - metadata:
      name: www
    spec:
      accessModes: [ "ReadWriteOnce" ]
      storageClassName: my-storage-class
      resources:
        requests:
          storage: 1Gi```

We should instead handle this in a way similar to the current `deduceResourceLimits` method in `AbstractKubernetesDeployer`.  `KubernetesDeployerProperties` should have some 'limits' and 'requests' values related to the volumeClaimTemplate which is used here...

```				new PersistentVolumeClaimBuilder().withNewSpec().withAccessModes(Arrays.asList("ReadWriteOnce"))
					.withNewResources().addToLimits(emptyStorageResource).addToRequests(emptyStorageResource)```

Also, the variable name `emptyStorageResource` isn't really a good name anymore since it won't be empty by default.

markpollack: I don't know what the minimum value can be ATM for the volume, but asking for 1 gig feels rather high considering the app won't be using the storage in the common SCDF use-case.

ilayaperumalg: I agree that the `memory` doesn't map directly to the `storage` usage here.

But, the resources `limits` and `requests` correspond to `cpu`, `memory` properties only as specified here: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/

It might be confusing to add an extra property such as `storage` inside the `limits` and `requests`. Maybe we can enforce an explicit property here? 

On the minimum value, I thought 1024Mi isn't too high anymore ;).  We can change it to a lower value for sure.

markpollack:
squashed and merged as 90d96b69afb36fd636f5faeacd77c8ec58bb9d00

Manually tested with @ilayaperumalg on minikube k8s version 1.9.  Will add 1.9 to  Acceptance Test infrastructure.

PR #189, opened 2018-02-13
Fix all deprecations

### LINKED ISSUE #183: Fix all deprecations
Should fix use of all deprecated methods whether those are external or internal. Especially we should not deprecate something internally and then not do anything about it.

chrisjs:
Fixes #183


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -375,31 +360,31 @@ public void setStatefulSet(
 	 */
 	@Deprecated
```
viniciusccarvalho: add `@deprecated Use {@link #getLimits()}` much like the one in `getCpu`?

chrisjs: yes its there already

PR #202, opened 2018-03-27
[Waiting for Feedback; do not merge] Shell style entrypoint can take command arguments as well

gitlabbin:
- check the PR: https://github.com/spring-cloud/spring-cloud-stream-app-maven-plugin/pull/15

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
@@ -120,10 +120,8 @@ public class DefaultContainerFactory implements ContainerFactory {
 			}
 			break;
 		case shell:
-			for (String key : request.getDefinition().getProperties().keySet()) {
-				String envVar = key.replace('.', '_').toUpperCase();
-				envVarsMap.put(envVar, request.getDefinition().getProperties().get(key));
-			}
+			appArgs = createCommandArgs(request);
+			appArgs.add(0, "--");
 			break;
 		}
 
--- a/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactoryTests.java
+++ b/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactoryTests.java
@@ -190,9 +190,10 @@ public class DefaultContainerFactoryTests {
 		Container containerShell = defaultContainerFactory.create("app-test",
 				appDeploymentRequestShell, null, false);
 		assertNotNull(containerShell);
-		assertTrue(containerShell.getEnv().get(0).getName().equals("FOO_BAR_BAZ"));
-		assertTrue(containerShell.getArgs().size() == 0);
-
+		assertTrue(containerShell.getArgs().get(0).equals("--"));
+		assertTrue(containerShell.getArgs().get(1).equals("--foo.bar.baz=test"));
+		assertTrue(containerShell.getArgs().size() == 2);
+		
 		props.put("spring.cloud.deployer.kubernetes.entryPointStyle", "exec");
 		AppDeploymentRequest appDeploymentRequestExec = new AppDeploymentRequest(definition,
 				resource, props);


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
```diff
@@ -120,10 +120,7 @@ public Container create(String appId, AppDeploymentRequest request, Integer port
 			}
 			break;
 		case shell:
-			for (String key : request.getDefinition().getProperties().keySet()) {
-				String envVar = key.replace('.', '_').toUpperCase();
-				envVarsMap.put(envVar, request.getDefinition().getProperties().get(key));
-			}
+			appArgs = createCommandArgs(request);
```
chrisjs: This change makes `shell` act like `exec` rather than adding anything additional. Can you clarify the use case here?

gitlabbin: yes, this change just make the same way for both `shell` and `exec`, once  the PR: spring-cloud/spring-cloud-stream-app-maven-plugin#15 merged.

sabbyanandan: Hi, @gitlabbin. We are unable to relate to what you're trying to achieve with this proposed change. You've pointed to a few links, but they do not describe a use-case or anything specific to what you've observed. Could you please explain _your_ use-case requirements? The more the details, the better it is for us to understand how the plugin and the deployer changes are related and necessary. 

This change, however, appears to collapse "exec" and "shell" with the same parsing behavior. Though it may work for your use-case, it violates the [EntryPointStyle](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/master/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/EntryPointStyle.java#L20-L23) contract. Let's start with the requirements and revisit it.

gitlabbin: Well, the initial PR https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/71
introduced three styles, as @trisberg described.

But the "EXEC" and "SHELL" should be handled as the same way.  
how it works PR https://github.com/spring-cloud/spring-cloud-stream-app-maven-plugin/pull/15  did that implementation

chrisjs: `exec` won't do shell variable replacement (ie: `$JAVA_OPTS` in the generated ENTRYPOINT), `shell` will. building an application using the maven plugin PR you submitted, I am able to change the entry point style to `shell` and have a resolved `$JAVA_OPTS` variable passed to my application, for example:

```
stream create --name yourstream --definition "source | sink"
stream deploy yourstream --properties "deployer.yourapp.kubernetes.entryPointStyle=shell,deployer.yourapp.kubernetes.environmentVariables=JAVA_OPTS=-Duser.timezone=Australia/Sydney"
```

given the above, it seems this issue can be closed? your use case for "shell" appears would be resolved in the maven plugin PR you submitted and then with applications built to use `shell` style, configuring properties on the k8s side would give you the ability to have env vars like `$JAVA_OPTS` resolved

gitlabbin: Great, @chrisjs , thanks for your test, but you just tested half 
- [x] env variables values can be changed by user
 - [ ] command line arguments can be changed by user

gitlabbin: yes or no,
- the existing common code already handle that
https://github.com/gitlabbin/spring-cloud-deployer-kubernetes/blob/2208ac31c78475019a8f824acb8bf3e7d2f43555/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java#L101
- now the change just keep the command line args, instead to do the bad conversion

chrisjs: correct - `spring.cloud.deployer.kubernetes.environmentVariables` first gets parsed in `getAppEnvironmentVariables`

then when entry point is shell the following is parsed into the env var map:

```
				for (String key : request.getDefinition().getProperties().keySet()) {
					String envVar = key.replace('.', '_').toUpperCase();
					envVarsMap.put(envVar, request.getDefinition().getProperties().get(key));
				}
```

which you have replaced with:

```appArgs = createCommandArgs(request);```

per the EntryPointStyle javadoc:

```
 * EntryPoint style used for a Docker image. Affects how app properties are passed in.
 * For exec app properties should be passed in as command line args, for shell they should
 * be passed in as environment variables and for boot they should be passed in using a
 * SPRING_APPLICATION_JSON environment variable.
```

when using `shell` how is parsing properties into env vars bad conversion?

Comment (gitlabbin) [reply src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java]: The SCDF design should be change, clearly it can't handle the use case.
```
export JAVA_OPTS=-Xmx1024M -Duser.timezone=xxxxxx -Djavax.net.ssl.trustStore=/usr/local/jvmkeystore
$JAVA_HOME/bin/java $JAVA_OPTS http_fetch_client.jar --contentUrl=https://www.tvnews.provider.com --user.token=xxx --user.passwd=xxxx
```

chrisjs:
Closing as the existing code would replace the functionality of `shell`.

PR #208, opened 2018-04-05
Add support to provide fabric8 client properties

### LINKED ISSUE #206: Add support for passing multiple Fabric8 Client connection properties
_From @sabbyanandan on March 13, 2018 17:23_

For each client instance, we need to support all the Fabric8 Client [properties](https://github.com/fabric8io/kubernetes-client#configuring-the-client). Consider using the `ConfigBuilder` to create the client for Skipper use.

_Copied from original issue: spring-cloud/spring-cloud-skipper#558_

ilayaperumalg:
- Add fabric8 client config properties as part of the k8s deployer properties under the prefix `fabric8`
 - Merge `namespace` property from deployer properties if not set in client properties already
 - Use these properties when creating k8s fabric8 client
    - Add a factory class that instantiates the client after merging the namespace property from deployer properties
 - Add test to verify the changes

Resolves #206

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAutoConfiguration.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesClientFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesClientFactory.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.deployer.spi.kubernetes;
 
+import io.fabric8.kubernetes.client.Config;
+import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClient;
 
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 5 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -40,6 +41,16 @@
 	 */
 	public static final String KUBERNETES_DEPLOYMENT_NODE_SELECTOR = "spring.cloud.deployer.kubernetes.deployment.nodeSelector";
 
+	private Config fabric8 = new Config();
```
ilayaperumalg: Will create a story to add documentation for all the available fabric8 configuration properties once this PR is accepted and merged.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAutoConfiguration.java
```diff
@@ -56,7 +56,7 @@ public TaskLauncher taskDeployer(KubernetesClient kubernetesClient,
 
 	@Bean
 	public KubernetesClient kubernetesClient() {
-		return new DefaultKubernetesClient().inNamespace(properties.getNamespace());
+		return KubernetesClientFactory.getKubernetesClient(this.properties);
```
chrisjs: Update copyright date

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesClientFactory.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.deployer.spi.kubernetes;
+
+import io.fabric8.kubernetes.client.Config;
+import io.fabric8.kubernetes.client.DefaultKubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClient;
+
+import org.springframework.beans.factory.annotation.Autowired;
```
chrisjs: from here down are all unused imports & update copyright date

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
chrisjs: `new Config()` is deprecated

ilayaperumalg: @chrisjs  Yes, I saw that when I had it in. The deprecation message says this: 
```
//In future releases (2.4.x) the public constructor will be empty.
//The current functionality will be provided by autoConfigure().
//This is a necessary change to allow us distinguish between auto configured values and builder values.
```
We'll use autoConfigure() once we switch to the updated client. Sounds ok?

chrisjs: per slack discussion, lets avoid calling deprecated methods

chrisjs:
merged in 7d1a1

PR #215, opened 2018-04-19
Update api version based on k8s version from cluster

### LINKED ISSUE #192: [K8S 1.9] Investigate the use of Statefulset and why it is mapped to "v1beta1"
More details [here](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/587aca8b6c88a97e20c9376e586c78f84ad96a60/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java#L104).

ilayaperumalg:
- The api version to use in some of the controller's REST mappings need to be updated
based on the k8s version in the cluster.
  - Check the k8s version using k8s http client and update the api version accordingly

Resolves #192


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesHttpClient.java
```diff
@@ -78,6 +83,26 @@ public Response delete(String resourceEndpoint, String appId) {
...
+	 * @return the appropriate API version to use
+	 */
+	public static String getApiVersionForK8sVersionFromCluster(KubernetesHttpClient client) {
+		try {
+			Response response = client.get("version", "");
+			Map<String, String> versionResponse = new ObjectMapper().readValue(response.body().string(), HashMap.class);
+			String k8sVersionFromCluster = versionResponse.get("gitVersion");
+			Version version = Version.valueOf(k8sVersionFromCluster.substring(1));
+			Version version110 = Version.valueOf("1.10.0");
+			return (version.greaterThanOrEqualTo(version110)) ? "v1" : "v1beta1";
```
chrisjs: when testing on both GKE 1.10.0 and 1.9.6 im getting `v1beta1` back from `KubernetesHttpClient.getApiVersionForK8sVersionFromCluster(..)`. seems like i should get `v1` when running against GKE 1.10.0.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesHttpClient.java
```diff
@@ -78,6 +83,26 @@ public Response delete(String resourceEndpoint, String appId) {
...
+	/**
+	 * Get API version to use for some of the Controller's API route mappings based on the kubernetes
+	 * version used in the underlying cluster.
+	 * @param client the KubernetesHttpClient to use
+	 * @return the appropriate API version to use
+	 */
+	public static String getApiVersionForK8sVersionFromCluster(KubernetesHttpClient client) {
+		try {
+			Response response = client.get("version", "");
+			Map<String, String> versionResponse = new ObjectMapper().readValue(response.body().string(), HashMap.class);
```
chrisjs: the unchecked assignment can be avoided by using say `new TypeReference<HashMap<String, String>>(){}` rather than `HashMap.class`.

trisberg:
Just be aware that if you change the API versions you might have to adjust any RBAC configurations to match this.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesHttpClient.java
ilayaperumalg: Weird, I tested this against both 1.9.x and 1.10.0 and saw the version updated correctly. will double check

ilayaperumalg: Verified it again and it looks ok to me. What is the response you get on `/version` endpoint via HTTP request?

ilayaperumalg:
@trisberg This PR only addresses the URL for statefulsets [here](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/master/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java#L102). 


I tried to deploy the existing configuration agains minikube pointing to the k8s 1.10.0 and it looks ok.

Do you think we need to update all the configurations that use `apiVersion: extensions/v1beta1` for versions greater than or equal to GKE 1.10.0?

chrisjs:
for context from our slack conversation, while stripping anything after `-` would work, feels we might be accounting just for the GKE case (i.e.: another provider could use another delimiter). since the major/minor version fields seem to differ in availability between platforms (and seems like those will be deprecated) they don't seem useful. the `gitVersion` field seems to always be available yet may contain additional characters depending on the platform, i.e.: `-gke.0`. we might want to consider trying to just extract a version matching `x.x.x` format from the `gitVersion` and use that along with any error reporting if we fail to find it. some test coverage would be great as well.

chrisjs:
i think some additional changes need to be made -- `KubernetesConfigurationPropertiesTests` is failing. looks related to how the `masterUrl` property is being set and then a subsequent call being made to check the api version, but nothing is listening on `localhost:8090`. 

also, the test property value is incorrect as it is does not have a trailing `/` and `KubernetesHttpClient.get(..)` as it stands will build a url such as `localhost:8090version`. seems like we may be able to do something simple like check the `resourceEndpoint` for a leading `/` and if not add it. if you want to make a fix to `KubernetesHttpClient.get(..)` as part of this that would be great, if not we should log an issue for it

ilayaperumalg:
@chrisjs yeah, I just see that after rebasing my PR. I will try to fix this as part of this PR.

ilayaperumalg:
Pushed changes to fix the `masterUrl` after rebasing.

PR #229, opened 2018-05-22
Add apiVersion to StatefulSet object

### LINKED ISSUE #222: Statefulsets seem broken 1.10
Statefulsets appear to be broken on 1.10. The following tests fail:

`KubernetesAppDeployerIntegrationTests.testDeploymentWithGroupAndIndex` and `KubernetesAppDeployerIntegrationTests.testMultipleInstancesDeploymentAndPartialState`.

When a pod is created it sits in `Pending` state forever. One interesting bit from the object description is it has no assigned node nor events.

child of #223

chrisjs:
Resolves #222


chrisjs:
The default `apiVersion` on the StatefulSet object is `apps/v1beta1`. The `apps/v1` endpoint is being used when the k8s target is 1.10+ so the `apiVersion` on the StatefulSet object needs to be set to `apps/v1`. 

Change tested on GKE 1.9 and 1.10

ilayaperumalg:
LGTM, merging.

PR #243, opened 2018-07-13
Use withPodManagementPolicy method rather than adding JSON attribute

### LINKED ISSUE #230: Review StatefulSet implementation
For example, In `KubernetesAppDeployer.createStatefulSet` we convert the built `StatefulSet` object to JSON, add a `podManagementPolicy` attribute then `POST` that JSON. Newer fabric8 clients look to have a `withPodManagementPolicy` method so this may be one potential cleanup.

chrisjs:
Resolves #230

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.deployer.spi.kubernetes;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.fabric8.kubernetes.api.model.Container;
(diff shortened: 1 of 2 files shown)


chrisjs:
this is probably the most cleanup we can do for now. 

ideally this would have included the removal of `KubernetesHttpClient` and related calls, replacing them with `apps().statefulSets().create(...)` / `delete(...)` from the fabric8 library but until they catch up with k8s 1.10, going this route would result in the `v1beta` endpoint being used which would essentially be going backwards from the change that was made to obtain the k8s version and choose `v1` if 1.10+ or `v1beta1` otherwise.

currently there is no auto detection of platform, but given the recent fabric8 work on catching up to 1.9 looks like there will be some sort of automatic version management via https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/BackwardsCompatibilityInterceptor.java which at that time we can remove the `KubernetesHttpClient` and related calls

ilayaperumalg:
LGTM, merging.

PR #262, opened 2019-02-14
Top-level property "namespace" is ignored when creating a new KubernetesClient

### LINKED ISSUE #261: Top-level property "namespace" is ignored when creating a new KubernetesClient
As a user, while attempting to override the "namespace" property, I notice that it doesn't influence the deployment to a different namespace. 

It turns out, in `KubernetesClientFactory`, we only take `fabric8.namespace` [property into account](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/master/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesClientFactory.java#L34), but the top-level `namespace` property is not considered in the decision. 

**Acceptance:**
Decide whether we still need both the properties. If yes, make `KubernetesClientFactory` to take both the properties into account when creating a new client.

Relates to https://github.com/spring-cloud/spring-cloud-dataflow/issues/2852#issuecomment-462957754.

chrisjs:
Resolves #261

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesClientFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesClientFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 the original author or authors.
+ * Copyright 2018-2019 the original author or authors.
  *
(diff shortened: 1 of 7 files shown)


ilayaperumalg:
LGTM, merging.

PR #295, opened 2019-04-10
Polish properties metadata

jvalkeal:
- As spring.cloud.deployer.kubernetes.requests and spring.cloud.deployer.kubernetes.limits
  groups use same type(Resource), boot can't accurately match grouping and thus it resolves
  name of spring.cloud.deployer.kubernetes.requests.cpu as spring.cloud.deployer.kubernetes.requests.cpu and
  name of spring.cloud.deployer.kubernetes.limits.cpu as cpu. Simply make separate class for these
  types to have different base type and to be able to configure fields.
- Add missing field javadocs to rest of a fields to have a description in a metadata.
- Change fabric8 to be @NestedConfigurationProperty so that we get real
  property names in a metadata instead of one bogus property name spring.cloud.deployer.kubernetes.fabric8
  which is actually a group.

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
@@ -24,6 +24,7 @@ import io.fabric8.kubernetes.api.model.VolumeMount;
 import io.fabric8.kubernetes.client.Config;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
 
 /**
  * @author Florian Rosenberg
@@ -42,8 +43,12 @@ public class KubernetesDeployerProperties {
 	 */
 	public static final String KUBERNETES_DEPLOYMENT_NODE_SELECTOR = "spring.cloud.deployer.kubernetes.deployment.nodeSelector";
 
+	/**
+	 * The maximum concurrent tasks allowed for this platform instance.
+	 */
 	private int maximumConcurrentTasks = 20;
 
+	@NestedConfigurationProperty
 	private Config fabric8 = Config.autoConfigure(null);
 
 	public Config getFabric8() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -42,8 +43,12 @@
 	 */
 	public static final String KUBERNETES_DEPLOYMENT_NODE_SELECTOR = "spring.cloud.deployer.kubernetes.deployment.nodeSelector";
 
+	/**
+	 * The maximum concurrent tasks allowed for this platform instance.
+	 */
 	private int maximumConcurrentTasks = 20;
 
+	@NestedConfigurationProperty
```
chrisjs: I've never been a big fan of exposing this, but understand there was a use-case for more "advanced users" to be able to tweak some knobs here. Understood its in the docs, people can see it etc, but do we really want to put these options out there in the UI, publishing them even more? Maybe possible just to ignore this? Maybe a question for @sabbyanandan

sabbyanandan: We talked about it yday. Most of the Fabric8 properties don't have an effect when there's only one platform, because Fabric8 properties get automatically bootstrapped in SCDF/Skipper when they get provisioned on K8s, which is much earlier in the lifecycle. And it is not overridable when deploying streams/tasks. 

However, when we have >1 platform, we will have >1 connection override possibilities also, so that's when someone can deploy stream/tasks with a different set of Fabric8 overrides. 

All that said, we thought of blacklisting the properties that don't apply as overrides, and likewise, it shouldn't show up in the dropdown in SCDF UI. I suggested blacklisting everything under Fabric8 (most of it are connection/auth props), but @markpollack wanted to check to see what can still apply, and allow them to be in the dropdown.

The blacklisting thing would be outside of the scope of this PR, though, a separate one possibly.

chrisjs: got it - ok. yah +1 to black/white list of those properties or maybe some wrapped around them like `namespace` - just feels awkward to expose the underlying fabric8 lib

jvalkeal:
@chrisjs Btw, feel free to update those property descriptions as those are now coming visible to UI as is. I kinda tried to add something so that we would not have empty description fields.

Screenshot from 2019-04-10 16-38-50

chrisjs:
merged in 58f1507390

PR #301, opened 2019-06-27
Allow nodeSelector to be set at the server global level

### LINKED ISSUE #300: Add support to apply nodeSelector as a global configuration for all the deployed apps
As an SCDF user, I'd like to have the ability to configure `nodeSelector` as a global setting (set in SCDF), so I can deploy and launch streams and tasks with `nodeSelector` criteria's globally.

Consider the following precedence order for the implementation.
- if the global property is set, it takes precedence by default 
- if the global property is set and also the deployment property is passed along, the deployment property _should_ take priority since the user here is trying to alter the behavior for that particular stream/task intentionally
- if nothing is set, do not alter any deployment behavior for the `nodeSelector`

When we have this ready, we could port the same behavior to https://github.com/spring-cloud/spring-cloud-scheduler-kubernetes/issues/25, too.

chrisjs:
Resolves #300

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -413,21 +413,31 @@ public class AbstractKubernetesDeployer {
 	/**
 	 * Get the nodeSelectors setting for the deployment request.
 	 *
-	 * @param properties The deployment request deployment properties.
+	 * @param deploymentProperties The deployment request deployment properties.
 	 * @return map of nodeSelectors
 	 */
-	protected Map<String, String> getNodeSelectors(Map<String, String> properties) {
+	private Map<String, String> getNodeSelectors(Map<String, String> deploymentProperties) {
 		Map<String, String> nodeSelectors = new HashMap<>();
 
-		String nodeSelectorsProperty = properties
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -413,21 +413,31 @@ protected boolean getHostNetwork(AppDeploymentRequest request) {
...
-		String nodeSelectorsProperty = properties
+		String nodeSelector = properties.getNodeSelector();
+
+		String nodeSelectorDeploymentProperty = deploymentProperties
 				.getOrDefault(KubernetesDeployerProperties.KUBERNETES_DEPLOYMENT_NODE_SELECTOR, "");
 
-		if (StringUtils.hasText(nodeSelectorsProperty)) {
-			String[] nodeSelectorPairs = nodeSelectorsProperty.split(",");
+		boolean hasGlobalNodeSelector = StringUtils.hasText(properties.getNodeSelector());
+		boolean hasDeployerPropertyNodeSelector = StringUtils.hasText(nodeSelectorDeploymentProperty);
```
sabbyanandan: `hasDeployerPropertyNodeSelector` -> is a bit misleading. It is not a deployer property, though, but rather it is a "per deployment-level" property ... so let's revise it.

PR #307, opened 2019-07-12
Add support for secretKeyRef

### LINKED ISSUE #264: Add secretKeyRef to the container for passing secrets to the streams/tasks
Currently we can pass environment variables to the tasks created but not as secret. In order to not expose sensible information, we should be able to support "secretKeyRef" as described in here https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables

chrisjs:
Resolves #264

Some testing notes:

can be tested on minikube. build deployer + skipper + scdf + images / set imagePullPolicy on both servers to `Never` to use local build

create a secret ie: put the following into `secret.yaml`:
```
apiVersion: v1
kind: Secret
metadata:
  name: testsecret
type: Opaque
data:
  password: dGhlcGFzc3dvcmQ=
```

`$ kubectl create -f secret.yaml`

in dataflow shell:

`app import --uri https://dataflow.spring.io/rabbitmq-docker-latest`

* for per-application setting:

```
stream deploy --name ticktock --properties "deployer.log.kubernetes.secretKeyRefs=[{envVarName: 'MY_SECRET', secretName: 'testsecret', dataKey: 'password'}]"
```

* for global setting edit the appropriate config map, for example add the following to skipper-config-rabbit:

```
                accounts:
                  default:
                    ....
                    secretKeyRefs:
                      - envVarName: MY_SECRET
                        secretName: testsecret
                        dataKey: password
```

then do:

`stream create --name ticktock --definition "time | log" --deploy`

* then verify the appropriate pod envs -- the specific pod if per app deployment, either pod if global:

`$ kubectl exec -it ticktock-log-v1-949b9c498-s9zh8 -- printenv | grep MY_SECRET`

replacing `ticktock-log-v1-949b9c498-s9zh8` with your log pod name, should result in:

```$ kubectl exec -it ticktock-log-v1-949b9c498-s9zh8  -- printenv | grep MY_SECRET
MY_SECRET=thepassword
```

* delete the secret:

`kubectl delete secret/testsecret`

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -20,7 +20,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Properties;
 import java.util.stream.Collectors;
 
 import io.fabric8.kubernetes.api.model.Container;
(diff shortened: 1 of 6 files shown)


ilayaperumalg:
LGTM, merging.

PR #308, opened 2019-07-15
Support ConfigMapKeyRef

### LINKED ISSUE #298: Add configmap to the container for passing configurations to the deployed streams/tasks
As a user, I'd like to have the option to optionally also pass `configmap` as a property to the deployed stream/task applications.

Similar to https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/264, we should investigate and decide whether we should make direct calls using the Fabric8 client or rely on spring-cloud-kubernetes instead. Eitherways, once when the feature is supported in the deployer, we would have to document the usage with an example in SCDF ref. guide.

Perhaps also revise the file-ingest ref. sample to include the `configmap` settings as opposed to passing the properties as inline overrides.

chrisjs:
Resolves #298

testing notes - pretty much same as secret key ref:

* build deployer, skipper, scdf

* create test CM, ie:

`kubectl create configmap testcm --from-literal=platform=k8s`

in dataflow shell:

`app import --uri https://dataflow.spring.io/rabbitmq-docker-latest`

* for per app deployment:

```
stream deploy --name ticktock --properties "deployer.log.kubernetes.configMapKeyRefs=[{envVarName: 'MY_CM', configMapName: 'testcm', dataKey: 'platform'}]"
```

* for global setting edit the appropriate config map, for example add the following to skipper-config-rabbit:

```
                accounts:
                  default:
                    ....
                    configMapKeyRefs:
                      - envVarName: MY_CM
                        configMapName: testcm
                        dataKey: platform
```

then deploy without props:

`stream create --name ticktock --definition "time | log" --deploy`

* in both cases:

then verify the appropriate pod envs -- the specific pod if per app deployment, either pod if global:

`$ kubectl exec -it ticktock-log-v1-949b9c498-s9zh8 -- printenv | grep MY_CM`

replacing `ticktock-log-v1-949b9c498-s9zh8` with your log pod name, should result in:

`MY_CM=k8s`

delete the CM:

`$ kubectl delete cm/testcm`

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.deployer.spi.kubernetes;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.fabric8.kubernetes.api.model.ConfigMapKeySelector;
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.ContainerBuilder;
 import io.fabric8.kubernetes.api.model.EnvVar;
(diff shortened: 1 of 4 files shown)


ilayaperumalg:
LGTM, verified on minikube, merging.

PR #311, opened 2019-07-18
Add ability to configure a basic pod security context

### LINKED ISSUE #306: Add support to configure Security Context for a Pod
As an SCDF user, I'd like to have the ability to deploy a stream/task pod with a specific user/group ID, so container processes should be running as the default root or a customized user/group ID [see: [docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)].

**Acceptance:**
- Add a new top-level deployer property to control `securityContext` at each deployment level
- Globally also, it should be possible to configure `securityContext` to all the deployed pods
- Add tests

chrisjs:
Resolves #306

some testing notes:

create for example a stream and set the `podSecurityContext` either per application or globally. the docs at: https://github.com/spring-cloud/spring-cloud-dataflow/pull/3354 show the variations.

to verify, the following can be done:

`kubectl get pod/ticktock-log-v1-6cfc4498f4-c5b8p -o yaml`

and look for the `securityContext` block, it should match the values set

or by inspecting the running process on the container (for long running ones like streams vs tasks), ie:

```
$ kubectl exec -it ticktock-log-v1-6cfc4498f4-c5b8p -- /bin/sh
$ ps -p 1 -o args,group,user
```

which would yield something like:

```
COMMAND                     GROUP    USER
java -jar /maven/log-sink-r nogroup  nobody
```

the UID/GID of 65534 resolves to `nobody` in this container.

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -25,11 +25,15 @@ import java.util.stream.Collectors;
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.PodList;
+import io.fabric8.kubernetes.api.model.PodSecurityContext;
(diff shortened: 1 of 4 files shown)


ilayaperumalg:
LGTM, verified and merging.

ilayaperumalg:
Rebased and merged as 6bc6109

PR #317, opened 2019-08-14
Move K8s scheduler into K8s deployer

### LINKED ISSUE #316: Move KubernetesScheduler to K8s-deployer

ilayaperumalg:
- Move K8s scheduler code changes spring-cloud-scheduler-kubernetes
 - Fix conflicting PowerMock and Mockito versions
 - Copy RelaxedNames from Spring Boot for EntryPointStyle

Resolves #316


src/main/java/org/springframework/cloud/deployer/scheduler/spi/kubernetes/EntryPointStyle.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.springframework.cloud.deployer.scheduler.spi.kubernetes.support.RelaxedNames;
+
+/**
+ * Defines container entry point styles that are available. The selected entry point style
+ * will determine how application properties are made available to the container.
+ *
+ * @author Chris Schaefer
+ */
+public enum EntryPointStyle {
```
chrisjs: we should probably use the existing `EntryPointStyle` rather than duping here as they will have the same options

src/main/java/org/springframework/cloud/deployer/scheduler/spi/kubernetes/ImagePullPolicy.java
```diff
@@ -0,0 +1,68 @@
...
+
+import org.springframework.cloud.deployer.scheduler.spi.kubernetes.support.RelaxedNames;
+
+/**
+ * Defines container image pull policies that are available. The selection of pull policy
+ * will determine when an image is pulled.
+ *
+ * @author Chris Schaefer
+ */
+public enum ImagePullPolicy {
```
chrisjs: we should probably use the existing `ImagePullPolicy` rather than duping here as they will have the same options

src/main/java/org/springframework/cloud/deployer/scheduler/spi/kubernetes/support/RelaxedNames.java
```diff
@@ -0,0 +1,258 @@
...
+
+import org.springframework.util.StringUtils;
+
+/**
+ * Generates relaxed name variations from a given source.
+ *
+ * @author Phillip Webb
+ * @author Dave Syer
+ */
+public final class RelaxedNames implements Iterable<String> {
```
chrisjs: i think it would be best not to copy classes over. after `RelaxedNames` was deprecated a comparable impl was made - see `org.springframework.cloud.deployer.spi.kubernetes.ImagePullPolicy.relaxedValueOf(..)`. this would also go away if we use `EntryPointStyle` and `ImagePullPolicy` as mentioned above rather than the dups

src/test/java/org/springframework/cloud/deployer/KubernetesTestSupport.java
```diff
@@ -14,20 +14,22 @@
  * limitations under the License.
  */
 
-package org.springframework.cloud.deployer.spi.kubernetes;
+package org.springframework.cloud.deployer;
```
chrisjs: any particular need for the package change here? the test check is still for "kubernetes" rather than something generic. if reverted, the protected->public changes below can go too

src/test/java/org/springframework/cloud/deployer/scheduler/spi/kubernetes/EntryPointStyleTests.java
```diff
@@ -0,0 +1,30 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link EntryPointStyle}
+ *
+ * @author Chris Schaefer
+ */
+public class EntryPointStyleTests {
```
chrisjs: this might shuffle elsewhere if we go with a single `EntryPointStyle` enum rather than dups

src/test/java/org/springframework/cloud/deployer/scheduler/spi/kubernetes/ImagePullPolicyTests.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link ImagePullPolicy}.
+ *
+ * @author Chris Schaefer
+ */
+public class ImagePullPolicyTests {
```
chrisjs: this might shuffle elsewhere if we go with a single `ImagePullPolicy` enum rather than dups

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployerIntegrationTests.java
```diff
@@ -54,6 +54,7 @@
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.cloud.deployer.KubernetesTestSupport;
```
chrisjs: unneeded if moved back under `spi.kubernetes` as mentioned above

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationTests.java
```diff
@@ -34,6 +34,7 @@
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.cloud.deployer.KubernetesTestSupport;
```
chrisjs: unneeded if moved back under spi.kubernetes as mentioned above

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherWithJobIntegrationTests.java
```diff
@@ -29,6 +29,7 @@
 import org.junit.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.cloud.deployer.KubernetesTestSupport;
```
chrisjs: unneeded if moved back under spi.kubernetes as mentioned above

src/main/java/org/springframework/cloud/deployer/scheduler/spi/kubernetes/ContainerCreator.java
```diff
@@ -0,0 +1,175 @@
...
+
+import org.springframework.cloud.deployer.scheduler.spi.core.ScheduleRequest;
+import org.springframework.util.Assert;
+
+/**
+ * Configures a {@link Container} that will be launched by a {@link CronJob}.
+ *
+ * @author Chris Schaefer
+ */
+class ContainerCreator {
```
chrisjs: most of this stuff already exists in `DefaultContainerFactory` - would be nice to reuse plus be able to take advantage of other functionality in there rather than having to dup it

jvalkeal:
I'd rather keep this and other PR's simple(as plain spi move) and do these other changes in a separate issues.

chrisjs:
> I'd rather keep this and other PR's simple(as plain spi move) and do these other changes in a separate issues.

i suppose that's fine as long as we commit to coming back and doing the cleanups before cutting a release. didn't see an issue for that so assumed it would all be wrapped up in here.

if we are doing a simple move, i think the following would still be valid requests as they are changes introduced rather than being moved over:

* https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/317/files#r313929180 - just remove the class copy and use the existing impls from:
  * https://github.com/spring-cloud/spring-cloud-scheduler-kubernetes/blob/master/src/main/java/org/springframework/cloud/scheduler/spi/kubernetes/EntryPointStyle.java#L58
  * https://github.com/spring-cloud/spring-cloud-scheduler-kubernetes/blob/master/src/main/java/org/springframework/cloud/scheduler/spi/kubernetes/ImagePullPolicy.java#L56

* https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/317/files#r313929978

* https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/317/files#r313933103

* https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/317/files#r313933275

* https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/317/files#r313933330

src/test/java/org/springframework/cloud/deployer/KubernetesTestSupport.java
ilayaperumalg: The idea was to use `KubernetesTestSupport` in both the deployer and scheduler tests.

jvalkeal:
Yeah those were good proposals. We should definitely commit to do the changes.

PR #318, opened 2019-08-15
Feature/k8s affinity

emedina:
Added support for affinity rules, both node and (anti-)pod.

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -516,7 +516,7 @@ public class AbstractKubernetesDeployer {
                     .buildPodAntiAffinity());
         } else if (StringUtils.hasText(podAntiAffinityProperty)) {
             KubernetesDeployerProperties podAntiAffinityProperties = PropertyParserUtils.bindProperties(request,
-                    "spring.cloud.deployer.kubernetes.affinity.odAntiAffinity", "odAntiAffinity");
+                    "spring.cloud.deployer.kubernetes.affinity.podAntiAffinity", "podAntiAffinity");
 
             affinity.setPodAntiAffinity(new AffinityBuilder()
                     .withPodAntiAffinity(podAntiAffinityProperties.getPodAntiAffinity())
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -16,13 +16,28 @@
 
 package org.springframework.cloud.deployer.spi.kubernetes;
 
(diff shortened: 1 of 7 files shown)


pivotal-issuemaster:
@emedina Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-deployer-kubernetes&pullRequestId=318)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-deployer-kubernetes&pullRequestId=318) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@emedina Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-deployer-kubernetes&pullRequestId=318)!

emedina:
Documentation updated as well in this PR:

https://github.com/spring-cloud/spring-cloud-dataflow/pull/3438

chrisjs:
thanks for doing both the impl and some docs!

two things id have to ask before looking at it further:

* please do not use wildcard imports
* a few files have been re-formatted, please undo that so we can see the changes clearly. if there is a need for files to be reformatted, they should be done in a separate commit

emedina:
All requests done and amended to the last commit.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.deployer.spi.kubernetes;
 
+import io.fabric8.kubernetes.api.model.Affinity;
```
chrisjs: these should be grouped with the other fabric8 imports

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -16,6 +16,9 @@
 
 package org.springframework.cloud.deployer.spi.kubernetes;
 
+import io.fabric8.kubernetes.api.model.NodeAffinity;
```
chrisjs: it would be preferred not to be exposing fabric8 types, rather create wrappers similar to `Tolerations`, `PodSecurityContext` in this class

emedina: But this is a complex type which requires several sub-types. Do you want me to just mimic it cloning the same structure?

chrisjs: ideally they would be wrapped. understood about the graph size. i merged the PR as-is and will revisit that issue at a later time. thanks!

chrisjs:
merged in 70aff2babfc659179c7a1dcaccd277687cf83e9e

PR #321, opened 2019-08-21
Give human readable errors when schedule name is invalid.

### LINKED ISSUE #320: Improve the error message for invalid cron expressions
Currently if a user attempts to create a schedule name that has characters (like capital letters) that are not accepted by kubernetes, our exit message is the default returned from fabric8 which is basically the regex expression ` ^a-z0-9([a-z0-9]*[a-z0-9])?$`.   We need to return a friendlier message.   More information is found here: https://github.com/spring-cloud/spring-cloud-dataflow/issues/3443

cppwfs:
resolves #320

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/scheduler/spi/kubernetes/KubernetesScheduler.java
+++ b/src/main/java/org/springframework/cloud/deployer/scheduler/spi/kubernetes/KubernetesScheduler.java
@@ -21,6 +21,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import io.fabric8.kubernetes.api.model.Container;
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/cloud/deployer/scheduler/spi/kubernetes/KubernetesScheduler.java
```diff
@@ -81,6 +85,16 @@ public void schedule(ScheduleRequest scheduleRequest) {
 		}
 	}
 
+	public void validateScheduleName(ScheduleRequest request) {
+		if(request.getScheduleName() == null) {
+			throw new IllegalArgumentException("The name for the schedule request is null");
```
chrisjs: why throw IllegalArgumentException in both places of this method vs CreateScheduleException? we throw that when the cron expression isnt valid as well

cppwfs: Good Catch

cppwfs: As it pertains to the cron expression, the exceptions we get back from k8's are pretty good.  
i.e.
```
Failed to create schedule CronJob.batch "bbb" is invalid: spec.schedule: Invalid value: "*/1 * * R C": Failed to parse int from R: strconv.Atoi: parsing "R": invalid syntax
``` 
when I tried to enter a cron expression of `*/1 * * R C`

chrisjs:
merged in 5304c785b6d1139c268441e2ec48cc65d3df6d76

PR #323, opened 2019-08-30
Refactor/cleanup scheduler SPI

### LINKED ISSUE #319: Scheduler merge cleanups
See https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/317 for cleanup comments to do post-merge (consolidate `ContainerCreator`, removed `RelaxedNames` class copy, etc)

we should do these cleanups before making a release of k8s deployer.

also, the package structure seems a bit awkward:

* org.springframework.cloud.deployer
  * scheduler.spi.kubernetes
  * spi.kubernetes

as a side note, it may be time to start looking at an overall architecture review for cleanups and making things a bit more cohesive

ilayaperumalg:
- Move scheduler packages into `org.springframework.cloud.deployer.spi.scheduler`
 - Remove RelaxedNames

Resolves #319

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,11 @@
 			<artifactId>kubernetes-client</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>com.squareup.okhttp3</groupId>
+			<artifactId>okhttp</artifactId>
+			<version>${okhttp3.version}</version>
+		</dependency>
+		<dependency>
 			<groupId>org.hashids</groupId>
 			<artifactId>hashids</artifactId>
 			<version>1.0.1</version>
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/EntryPointStyle.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/EntryPointStyle.java
@@ -16,9 +16,12 @@
 
 package org.springframework.cloud.deployer.spi.kubernetes;
 
-import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.Map;
 
-import org.springframework.cloud.deployer.scheduler.spi.kubernetes.support.RelaxedNames;
+import org.springframework.boot.context.properties.bind.Bindable;
+import org.springframework.boot.context.properties.bind.Binder;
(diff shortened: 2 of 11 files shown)


chrisjs:
when things were copied from scheduler and merged in https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/commit/878104dfcafcf943d980a69ee327fc864f3fecf7 it appears `KubernetesSchedulerTests` was renamed to `KubernetesSchedulerTestsScheduler`

when running a build + the test suite, this test class was not run by surefire due to that

renaming the class back to `KubernetesSchedulerTests` will then run as part of the maven test build, and there are failures

jvalkeal:
Right, I'm pushing a change to see what CI thinks.

jvalkeal:
@ilayaperumalg ah can't rename as other PR's are in

jvalkeal:
Merged per d404e69809ca2c51aeac1f00eedd6b60d5314033

chrisjs:
good to see its green, tho the need to set the `okhttp` version seems odd, for example we don't do that in the "old" scheduler project - https://github.com/spring-cloud/spring-cloud-scheduler-kubernetes/blob/master/pom.xml

`mvn dependency:tree` on `spring-cloud-scheduler-kubernetes` master yields:

```
[INFO] org.springframework.cloud:spring-cloud-scheduler-kubernetes:jar:2.0.1.BUILD-SNAPSHOT
[INFO] +- io.fabric8:kubernetes-client:jar:4.1.0:compile
[INFO] |  +- io.fabric8:kubernetes-model:jar:4.1.0:compile
[INFO] |  |  +- com.fasterxml.jackson.module:jackson-module-jaxb-annotations:jar:2.9.9:compile
[INFO] |  |  \- javax.validation:validation-api:jar:2.0.1.Final:compile
[INFO] |  +- com.squareup.okhttp3:okhttp:jar:3.9.1:compile
[INFO] |  |  \- com.squareup.okio:okio:jar:1.13.0:compile
[INFO] |  +- com.squareup.okhttp3:logging-interceptor:jar:3.9.1:compile
```

which appears to bring in the correct deps without any explicit setting. maybe im missing something obvious here?

PR #327, opened 2019-10-21
Allow setting of statefulset init container image

### LINKED ISSUE #315: Supports use of images other than busybox in KubernetesAppDeployer
I am trying to use Spring Cloud Data Flow on internal installation of K8S in our company.

Unforunately, I cannot use the 'busybox' image according to the security policy of our company.
The security team provides white list of base images permitted to use in the internal K8S cluster.

'busyboyx' is hard coded in [KubernetesAppDeployer.java](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/master/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java#L440)

If it is possible to specify other image name, it will be very helpful for us.

chrisjs:
Resolves https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/315

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
@@ -522,6 +522,11 @@ public class KubernetesDeployerProperties {
 	 */
 	private PodAntiAffinity podAntiAffinity;
 
+	/**
+	 * A custom init container image name to use when creating a StatefulSet
+	 */
+	private String statefulSetInitContainerImageName;
+
 	public String getNamespace() {
 		return namespace;
 	}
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -522,6 +522,11 @@ public Long getFsGroup() {
 	 */
 	private PodAntiAffinity podAntiAffinity;
 
+	/**
+	 * A custom init container image name to use when creating a StatefulSet
+	 */
+	private String statefulSetInitContainerImageName;
```
tzolov: Shouldn't this default to "busybox" as well?

chrisjs: the intent is to only set it if you want to specify a custom image name. it could be defaulted here instead tho, i can go either way..

tzolov:
rebased and merged at 5279e6cc6c2c7633f471c9f4bdb710531925b1df

benelog:
@chrisjs 
Thank you for resolving it!

Issue #328, opened 2019-10-25
Fix configuration properties with kebab-case 


ilayaperumalg:
With Spring Boot 2.x supporting the kebab-case (ex: my-property-key) only if the `@ConfigurationProperties` is annotated with a prefix in kebab-case or in the `application.properties`, we start to some of the properties such as the deployer/app properties properties are not applied correctly to the target platform.

For instance, from the SCDF UI, we show the deployer properties with the kebab-case (ex: pod-annotations) and once user configures these properties (ex: deployer.<myapp>.kubernetes.pod-annotations=key:value), this deployer property doesn't get applied at all.

We need to:

 - Fix the places wherever the kebab-case properties are omitted
 - Update the documentation to replace the existing kebab-case properties

chrisjs:
as it stands, pretty sure 100% of deployment properties that are plucked out of the deployment properties map are not in "kebab-case". how properties as a whole should likely be reviewed/revamped as there are many variations on how they are mapped. some are mapped into boot configprops on server start up, or that class is used as a binding for more complex objects, plucking app specific things out of deployment properties map, etc. also there are some precedence rules between "global" and "app" specific properties. 

so a bit to think about when the time comes.

PR #330, opened 2019-10-28
Add basic support for custom init containers

### LINKED ISSUE #187: Support InitContainer
I have a requirement to support `InitContainer` in pods and jobs, so I was thinking of adding `initResource` property to `AppDeploymentRequest` and then check if that is set `createPodSpec` we add it.

Is that OK as we are adding property to `org.springframework.cloud.deployer.spi.core` or I should only work with `spring-cloud-deployer-kubernetes`

chrisjs:
Resolves #187

---

some manual testing notes that can be done on minikube:

* build & install deployer
* build skipper & image locally
* build scdf & image locally
* change `ImagePullPolicy` in both scdf & skipper deployment to `Never`
* in scdf shell:

```
stream deploy --name ticktock --properties "deployer.log.kubernetes.initContainer={containerName: 'test', imageName: 'busybox:latest', commands: ['sh', '-c', 'echo hello']}"
```

* verify init container is present in pod:
`$ kubectl describe log-pod-name`

* verify under `Init Containers` there is a container called `test`
* there should be no errors, state `Terminated`, exit code 0

* verify command was executed:
```
$ kubectl logs log-pod-name -c test
hello
$
```

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -26,6 +26,7 @@ import java.util.stream.Collectors;
 import io.fabric8.kubernetes.api.model.Affinity;
 import io.fabric8.kubernetes.api.model.AffinityBuilder;
 import io.fabric8.kubernetes.api.model.Container;
+import io.fabric8.kubernetes.api.model.ContainerBuilder;
 import io.fabric8.kubernetes.api.model.ContainerStatus;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.PodList;
@@ -224,6 +225,8 @@ public class AbstractKubernetesDeployer {
 		setPodSecurityContext(request, podSpec);
 
 		setAffinityRules(request, podSpec);
+
+		setInitContainer(request, podSpec);
 
 		return podSpec.build();
 	}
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
LGTM, will merge after testing.

ilayaperumalg:
Looks like the newly added deployer properties work only if we pass the properties this way:

```
deployer.log.kubernetes.initContainer={containerName: 'test', imageName: 'busybox:latest', commands: ['sh', '-c', 'echo hello']}
```
If we choose to pass these properties from the Data Flow UI (even after modifying the kebab case to camel case ex: init-container to initContainer), the properties don't get applied. This may be related to how to retrieve the properties of a specific Type (InitContainer in this case).

 @chrisjs Could you give this a try from the Data Flow UI and see if you see the same issue?

sabbyanandan:
There have been many reports of this case related questions lately. Also, all the reports come from the SCDF Dashboard users.

The code is currently looking specifically for the `initContainer `, and that's strictly a requirement for certain deployer properties. In this case, here's the snippet that expects the said format: https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/330/files#diff-e6a1f24b00f235994838c97c63fd33a5R573.

We have to revise deployer properties eventually (not in 2.3 timeline, however -- it could be perceived as a breaking change). For the time being, we will have to review why we are extracting the properties to the kebab case, which is then used in the Dashboard, and that causes the confusion.

@jvalkeal would have insights to share.

sabbyanandan:
Here are some of the hits that I was referring to in the previous comment.

1) https://gitter.im/spring-cloud/spring-cloud-dataflow?at=5daf23f09825bd6bacab231e
2) https://stackoverflow.com/questions/58523869/spring-data-flow-iam-role-assignment-to-pods-using-pod-annotations
3) https://github.com/spring-cloud/spring-cloud-dataflow/issues/3578

All of them use the SCDF Dashboard, of course.

chrisjs:
"kabab case" isn't going to work when pulling out of a map. if we need to change the way deployer properties are handled, we should probably do them under the other issue you submitted in one sweep with some design before hand. id prefer not just to do this as a one off

ilayaperumalg:
Merging this PR as the issues mentioned in this PR can be addressed via https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/328

ilayaperumalg:
Merged as 3397aec

PR #333, opened 2019-11-04
Fix K8s deployer properties retrieval

### LINKED ISSUE #328: Fix configuration properties with kebab-case

ilayaperumalg:
- Some of the K8s deployer properties which have the fixed camel case property names need to be parsed explicitly to allow retrieving the other supported property name keys.
   - Add a utility to parse the camel case property names and retrieve the property values even if any of the supported property name syntax is used
   - Move RelaxedNames util class as it helps find the possible property name combinations
       - The reason to go with this class is simply to fetch all possible combination of property name values and this is different from how property binding works in Spring Boot 2
   - Add tests to showcase the issue

Resolves #328

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -67,6 +67,7 @@ import static java.lang.String.format;
  * @author David Turanski
  * @author Chris Schaefer
  * @author Enrique Medina Montenegro
+ * @author Ilayaperumal Gopinathan
  */
 public class AbstractKubernetesDeployer {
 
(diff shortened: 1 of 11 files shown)


chrisjs:
merged in 2f2b701e8d3e205a3e9c23a67ff1e28d197c4610

PR #339, opened 2019-11-20
Add way to set LoadBalancerIP when deploying a LoadBalancer service.

### LINKED ISSUE #338: Allow definition of loadBalancerIP
Hi everyone,
when I deploy an application via SCDF the problem is that the service name changes due to the version being appended. Because I would like to have a consistent way to address my services, it would be great if I could give the LoadBalancer service a static cluster IP. 
I created a simple PR to add this but I'm not sure how to test this properly. https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/339

rbraeunlich:
Resolves #338


chrisjs:
closed per comment: https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/338#issuecomment-557089512

PR #354, opened 2019-12-12
Meaningful message returned if schedule name too long

### LINKED ISSUE #353: Need a meaningful error message when schedule names are too long
Currently the deployer returns the following exceptions if your schedule name exceeds the maximum char size for the field.
When creating a container the constraint reports `must be between 0 and 63`. 
In the last version a KubernetesClientException was fired.   In this case reported that the Schedule  was not created but no reason why.   
We need to let the user know that their schedule name exceeded maximum allowed characters.   

This issue is a child of https://github.com/spring-cloud/spring-cloud-dataflow/issues/3682

cppwfs:
resolves #353

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/scheduler/kubernetes/KubernetesScheduler.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/scheduler/kubernetes/KubernetesScheduler.java
@@ -32,6 +32,7 @@ import io.fabric8.kubernetes.api.model.batch.CronJobBuilder;
 import io.fabric8.kubernetes.api.model.batch.CronJobList;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClientException;
+import javax.validation.ConstraintViolationException;
 
 import org.springframework.cloud.deployer.spi.scheduler.CreateScheduleException;
 import org.springframework.cloud.deployer.spi.scheduler.ScheduleInfo;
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/cloud/deployer/spi/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -190,4 +210,12 @@ private void setImagePullSecret(ScheduleRequest scheduleRequest, CronJob cronJob
 					.add(localObjectReference);
 		}
 	}
+
+	private boolean isScheduleNameTooLong(String message ) {
```
chrisjs: maybe `validationScheduleNameLength` or similar -- mainly to match something liek the other `validateScheduleName`

src/main/java/org/springframework/cloud/deployer/spi/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -190,4 +210,12 @@ private void setImagePullSecret(ScheduleRequest scheduleRequest, CronJob cronJob
...
+		boolean result = false;
+		if(StringUtils.hasText(message) && message.contains("must be no more than") && message.endsWith("characters")) {
```
chrisjs: consider just inlining

cppwfs:
Submitted issue https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/356 so that we can investigate using the Constraints provided by Fabric8 instead of home-rolling our own.

chrisjs:
merged in 034f9c426a8f512d77defcea996f5bc440c31052

PR #366, opened 2020-02-13
Convert cmdline args as environment props in Shell entrypoint

### LINKED ISSUE #365: Add option to pass "executionId" and "platformName" either as cmd-line arguments or plain environment variables
@cppwfs commented on [Thu Nov 21 2019](https://github.com/spring-cloud/spring-cloud-dataflow/issues/3651)

Currently the executionId and platformName are passed to the task to be launched via the command line args.  i.e. 
```
--spring.cloud.dataflow.platform_name: default
--spring.cloud.task.executionid: 91
``` 
If a user uses the shell entry point for launching tasks on kubernetes these values can't be passed as command line args (they are ignored by the entrypoint).   We need to have a  boolean flag `passDefaultExecutionArgsAsProperties` in data flow that allows users to set this flag to `true` if they want these values passed as properties.  It defaults to false.

**NOTE:**
In the docs, further clarify that when `shell` option is selected, the command line arguments from the deployment request _are not converted_ into environment variables.

ilayaperumalg:
- When the docker entrypoint style is set to `shell`, the commandline arguments are ignored.
To fix this issue, convert all the commandline arguments and set as environment properties.

When applying the commandline arguments, we also need to make sure the commandline arguments override the existing app properties.

This will also make sure some of the platform specific properties to be passed in as environment properties.

 - Add tests

Resolves #365

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
@@ -16,30 +16,6 @@
 
 package org.springframework.cloud.deployer.spi.kubernetes;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import io.fabric8.kubernetes.api.model.ConfigMapKeySelector;
-import io.fabric8.kubernetes.api.model.Container;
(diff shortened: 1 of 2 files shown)


cppwfs:
LGTM

chrisjs:
given this changes how the functionality was previously defined, the docs should be updated as well: https://docs.spring.io/spring-cloud-dataflow/docs/current-SNAPSHOT/reference/htmlsingle/#_entry_point_style

PR #373, opened 2020-03-11
Expose Job specific properties

### LINKED ISSUE #332: BackoffLimit will not be set
**Kubernetes Pod backoff failure policy**

From the [k8s documentation][1]:
> There are situations where you want to fail a Job after some amount of retries due to a logical error in configuration etc. To do so, set .spec.backoffLimit to specify the number of retries before considering a Job as failed. The back-off limit is set by default to 6.

**Spring cloud dataflow:**

When a job has failed, we actually don't want a retry. In other words, we want to set the `backoffLimit: 1` in our Sprint Cloud Dataflow config file.

We have tried to set it like the following: 

    deployer.kubernetes.spec.backoffLimit: 1

or even

    deployer.kubernetes.backoffLimit: 1

But both is not transmitted to our Kubernetes Cluster.

After 6 tries, we see the following message:

> status:   conditions:
>     - lastProbeTime: '2019-10-22T17:45:46Z'
>       lastTransitionTime: '2019-10-22T17:45:46Z'
>       message: Job has reached the specified backoff limit
>       reason: BackoffLimitExceeded
>       status: 'True'
>       type: Failed   failed: 6   startTime: '2019-10-22T17:33:01Z'

Actually we want to fail fast (1 or 2 tries maximum)

Question: How can we properly set this property, so that all task triggered by SCDF will fail maximum once on Kubernetes?

We have also tried the property:

    deployer:
       kubernetes:
          maxCrashLoopBackOffRestarts: Never # No retry for failed tasks

But the jobs are still failing 6 times instead of 1.

For completeness sake:

 1. I am scheduling a task in SCDF
 2. The task is triggered on Kubernetes (more specifically Openshift)
 3. When I check the configuration on the K8s-platform, I see that it still has a backoffLimit of 6, instead of 1:

YAML config snippet taken from the running pod:

      spec:
        backoffLimit: 6
        completions: 1
        parallelism: 1

In the official [documentation][2], it says: 

    `maxCrashLoopBackOffRestarts` - Maximum allowed restarts for app that is in a CrashLoopBackOff. Values are `Always`, `IfNotPresent`, `Never`

But `maxCrashLoopBackOffRestarts` takes an integer. So I guess the documentation is not accurate.

The pod is then restarted 6 times.

I have tried to set those properties unsuccessfully:

    spring.cloud.dataflow.task.platform.kubernetes.accounts.defaults.maxCrashLoopBackOffRestarts: 0
    spring.cloud.deployer.kubernetes.maxCrashLoopBackOffRestarts: 0
    spring.cloud.scheduler.kubernetes.maxCrashLoopBackOffRestarts: 0

None of those has worked.

Any idea?

  [1]: https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/
  [2]: https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#_tasks

### LINKED ISSUE #337: Expose Job specific Kubernetes Properties in Deployer
Currently there is no way to cleanup the pods and Jobs created by the Spring Cloud Tasks automatically.  

Based on the below documentation, we could see there are configs available for Jobs to cleanup the finished Jobs and pods automatically. 
https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

_**Useful configs for Jobs cleanup:**_

1.  **.spec.ttlSecondsAfterFinished**    - Useful for specifying the time to live seconds for Jobs after it is finished.
2. **.spec.backoffLimit** - useful for specify the number of retries before considering a Job as failed. Default backoff Limit is 6.Addition of this property will help us set the limit to 1 or 2 at the max.
3. **.spec.template.spec.restartPolicy** - useful for specifying the restart policy. default seem to be "Always". Addition of this property will help us set the restart policy to "Never"

These configs are not currently exposed in deployer as per the below documentation.
https://docs.spring.io/spring-cloud-dataflow/docs/2.2.1.RELEASE/reference/htmlsingle/#configuration-kubernetes-deployer

Since there is no way to cleanup the pods and Jobs created automatically, Exposing these properties will really benefit everyone.

ilayaperumalg:
- Add RestartPolicy and BackoffLimit to Pod/Job spec respectively.
 - Add KubernetesTaskLauncherProperties to add these new properties as they are applicable only to task.
 - Update tests

Resolves #337
Resolves #332

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 the original author or authors.
+ * Copyright 2015-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
@@ -19,12 +19,12 @@ package org.springframework.cloud.deployer.spi.kubernetes;
 import java.util.ArrayList;
 import java.util.List;
 
-import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.api.model.NodeAffinity;
 import io.fabric8.kubernetes.api.model.PodAffinity;
 import io.fabric8.kubernetes.api.model.PodAntiAffinity;
 import io.fabric8.kubernetes.api.model.Volume;
 import io.fabric8.kubernetes.api.model.VolumeMount;
+import io.fabric8.kubernetes.client.Config;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.NestedConfigurationProperty;
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
@@ -72,8 +72,7 @@ public class KubernetesTaskLauncher extends AbstractKubernetesDeployer implement
 
 	@Autowired
 	public KubernetesTaskLauncher(KubernetesDeployerProperties deployerProperties,
-			KubernetesTaskLauncherProperties taskLauncherProperties,
-	                             KubernetesClient client) {
+			KubernetesTaskLauncherProperties taskLauncherProperties, KubernetesClient client) {
 		this(deployerProperties, taskLauncherProperties, client, new DefaultContainerFactory(deployerProperties));
 	}
 
@@ -116,18 +115,14 @@ public class KubernetesTaskLauncher extends AbstractKubernetesDeployer implement
 
 			Map<String, String> jobAnnotations = getJobAnnotations(request);
 
-			Map<String, String> jobSpecProperties = new HashMap<>();
-
-			Integer backOffLimit = getBackoffLimit(request);
-
-			if (backOffLimit != null) {
-				jobSpecProperties.put(BACKOFF_LIMIT_KEY, String.valueOf(backOffLimit));
-			}
-
 			podSpec.setRestartPolicy(getRestartPolicy(request).name());
 
 			if (this.properties.isCreateJob()) {
-				launchJob(appId, podSpec, podLabelMap, idMap, jobAnnotations, jobSpecProperties);
+				Integer backOffLimit = getBackoffLimit(request);
+				launchJob(appId, podSpec, podLabelMap, idMap, jobAnnotations,
+						(backOffLimit != null) ?
+								Collections.singletonMap(BACKOFF_LIMIT_KEY, String.valueOf(backOffLimit)) :
+								Collections.EMPTY_MAP);
 			}
 			else {
 				launchPod(appId, podSpec, podLabelMap, idMap, jobAnnotations);
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -19,12 +19,12 @@
 import java.util.ArrayList;
 import java.util.List;
 
-import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.api.model.NodeAffinity;
 import io.fabric8.kubernetes.api.model.PodAffinity;
 import io.fabric8.kubernetes.api.model.PodAntiAffinity;
 import io.fabric8.kubernetes.api.model.Volume;
 import io.fabric8.kubernetes.api.model.VolumeMount;
+import io.fabric8.kubernetes.client.Config;
```
chrisjs: copyright date should be bumped

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherProperties.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	public Integer getBackoffLimit() {
+		return backoffLimit;
+	}
+
+	public void setBackoffLimit(Integer backoffLimit) {
+		this.backoffLimit = backoffLimit;
+	}
+
```
chrisjs: can remove empty lines

src/main/java/org/springframework/cloud/deployer/spi/scheduler/kubernetes/KubernetesSchedulerProperties.java
```diff
@@ -19,6 +19,7 @@
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.deployer.spi.kubernetes.EntryPointStyle;
 import org.springframework.cloud.deployer.spi.kubernetes.ImagePullPolicy;
+import org.springframework.cloud.deployer.spi.kubernetes.RestartPolicy;
```
chrisjs: copyright date update

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherWithJobIntegrationTests.java
```diff
@@ -147,4 +153,91 @@ public void testJobAnnotations() {
...
+
+		String taskName = request.getDefinition().getName();
+
+		log.info("Checking job spec of {}...", taskName);
+
+		List<io.fabric8.kubernetes.api.model.batch.Job> jobs = kubernetesClient.batch().jobs().withLabel("task-name", taskName).list().getItems();
+
+		assertThat(jobs.size(), is(1));
+
+		io.fabric8.kubernetes.api.model.batch.Job job = jobs.get(0);
```
chrisjs: class is already imported, doesn't need to be fully qualified

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherWithJobIntegrationTests.java
```diff
@@ -147,4 +153,91 @@ public void testJobAnnotations() {
...
+
+		assertThat(launchId, eventually(hasStatusThat(
+				Matchers.<TaskStatus>hasProperty("state", Matchers.is(LaunchState.launching))), timeout.maxAttempts,
+				timeout.pause));
+
+		String taskName = request.getDefinition().getName();
+
+		log.info("Checking job spec of {}...", taskName);
+
+		List<io.fabric8.kubernetes.api.model.batch.Job> jobs = kubernetesClient.batch().jobs().withLabel("task-name", taskName).list().getItems();
```
chrisjs: class is already imported, doesn't need to be fully qualified

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherWithJobIntegrationTests.java
```diff
@@ -147,4 +153,91 @@ public void testJobAnnotations() {
...
+
+		assertThat(jobs.size(), is(1));
+
+		io.fabric8.kubernetes.api.model.batch.Job job = jobs.get(0);
+		assertTrue(job.getSpec().getBackoffLimit().equals(5));
```
chrisjs: could just use `assertEquals` here (or assertThat to be consistent with the one above)

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherWithJobIntegrationTests.java
```diff
@@ -147,4 +153,91 @@ public void testJobAnnotations() {
...
+
+
+		log.info("Checking job pod spec annotations of {}...", taskName);
```
chrisjs: looks like this may be copy/paste err?

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherWithJobIntegrationTests.java
```diff
@@ -147,4 +153,91 @@ public void testJobAnnotations() {
...
+
+		List<Pod> pods = kubernetesClient.pods().withLabel("task-name", taskName).list().getItems();
+
+		assertThat(pods.size(), is(1));
+
+		Pod pod = pods.get(0);
```
chrisjs: `pod` is ununsed

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherWithJobIntegrationTests.java
```diff
@@ -147,4 +153,91 @@ public void testJobAnnotations() {
...
+
+		assertThat(taskName, eventually(hasStatusThat(
+				Matchers.<TaskStatus>hasProperty("state", Matchers.is(LaunchState.unknown))), timeout.maxAttempts,
+				timeout.pause));
+	}
+
+
+	@Test
+	public void testJobSpecWithInvalidRestartPolicy() {
+		log.info("Testing {}...", "JobSpecProperties");
```
chrisjs: should be `JobSpecWithInvalidRestartPolicy`

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherWithJobIntegrationTests.java
```diff
@@ -147,4 +153,91 @@ public void testJobAnnotations() {
...
+		Resource resource = testApplication();
+		Map<String, String> appRequest = new HashMap<>();
+		appRequest.put("spring.cloud.deployer.kubernetes.restartPolicy", "Always");
+		appRequest.put("spring.cloud.deployer.kubernetes.backoffLimit", "5");
+		AppDeploymentRequest request = new AppDeploymentRequest(definition, resource, appRequest);
+
+		log.info("Launching {}...", request.getDefinition().getName());
+
+		try {
+			String launchId = kubernetesTaskLauncher.launch(request);
```
chrisjs: `launchId` is unused

src/test/java/org/springframework/cloud/deployer/spi/scheduler/kubernetes/KubernetesSchedulerPropertiesTests.java
```diff
@@ -25,6 +25,7 @@
 
 import org.springframework.cloud.deployer.spi.kubernetes.EntryPointStyle;
 import org.springframework.cloud.deployer.spi.kubernetes.ImagePullPolicy;
+import org.springframework.cloud.deployer.spi.kubernetes.RestartPolicy;
```
chrisjs: copyright date update

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
```diff
@@ -61,19 +62,27 @@
...
+			KubernetesTaskLauncherProperties taskLauncherProperties,
 	                             KubernetesClient client) {
-		this(properties, client, new DefaultContainerFactory(properties));
+		this(deployerProperties, taskLauncherProperties, client, new DefaultContainerFactory(deployerProperties));
 	}
 
 	@Autowired
-	public KubernetesTaskLauncher(KubernetesDeployerProperties properties,
+	public KubernetesTaskLauncher(KubernetesDeployerProperties kubernetesDeployerProperties,
+			KubernetesTaskLauncherProperties taskLauncherProperties,
```
chrisjs: formatting off

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
```diff
@@ -107,8 +116,18 @@ public String launch(AppDeploymentRequest request) {
 
 			Map<String, String> jobAnnotations = getJobAnnotations(request);
 
-			if (properties.isCreateJob()) {
-				launchJob(appId, podSpec, podLabelMap, idMap, jobAnnotations);
+			Map<String, String> jobSpecProperties = new HashMap<>();
```
chrisjs: given this only applies to `Job`'s, can just do this in `launchJob` passing in the request rather than building the map, etc

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
```diff
@@ -235,14 +254,18 @@ private void launchPod(String appId, PodSpec podSpec, Map<String, String> labelM
...
-						   Map<String, String> annotations) {
+						   Map<String, String> annotations, Map<String, String> jobSpecProperties) {
 		ObjectMeta objectMeta = new ObjectMetaBuilder().withLabels(podLabelMap).addToLabels(idMap).build();
 		PodTemplateSpec podTemplateSpec = new PodTemplateSpec(objectMeta, podSpec);
 
 		JobSpec jobSpec = new JobSpecBuilder()
 				.withTemplate(podTemplateSpec)
 				.build();
 
+		if (jobSpecProperties.containsKey(BACKOFF_LIMIT_KEY)) {
```
chrisjs: per above could just get this from `getBackoffLimit(request)`

ilayaperumalg: The `request` is not available in this method and also I thought it is not necessary to pass the entire `request` down here.

chrisjs: right; would need to be passed in. regardless, i think we should consider consolidating the signatures of `launchJob` and `launchPod` otherwise we will keep tacking on args. a value object could carry say the first 4 args of both methods, then if request is embedded, we wouldn't have to tack on jobAnnotations, this new map, etc

ilayaperumalg: Yeah, it makes sense. I have them addressed and updated the PR. Thanks!

chrisjs:
merged in 86ef8ddb354d50b78804d4aa7147c2aa85518d66

PR #380, opened 2020-04-16
Extend scheduler properties to support available deployer properties

### LINKED ISSUE #331: Support deployment options in Scheduler CronJob's
CronJob's in the merged Scheduler API should be able to take advantage of deployment options like volume mounts and so on. This was left out during the consolidation. Leaving this open ended for now until scope is understood.

ilayaperumalg:
- The KubernetesSchedulerProperties now extends from the KubernetesDeployer properties
 - The AppDeployer/TaskLauncher continue to use the KubernetesDeployer properties while the KubernetesScheduler is setup to use the Scheduler properties which extend the Deployer properties as well
 - Create a common Deployer properties resolver to resolve the properties when processing the PodSpec for app/task/scheduler
 - Refactor the existing code to avoid duplicates
 - Update tests

Resolves #331


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/support/PropertyParserUtils.java
```diff
@@ -69,7 +71,7 @@
 	 * @param yamlLabel the label representing the field to bind to
 	 * @return a {@link KubernetesDeployerProperties} with the bound property data
 	 */
-	static KubernetesDeployerProperties bindProperties(AppDeploymentRequest request, String propertyKey, String yamlLabel) {
+	public static KubernetesDeployerProperties bindProperties(AppDeploymentRequest request, String propertyKey, String yamlLabel) {
```
cppwfs, line 74: We can remove this method, it is not being used AFAIC.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DeploymentPropertiesResolver.java
```diff
@@ -0,0 +1,801 @@
...
+
+		return  (this.properties instanceof KubernetesSchedulerProperties) ?
+				((KubernetesSchedulerProperties) this.properties).getTaskServiceAccountName() : null;
+	}
+
+	protected static String getPropertyValue(Map<String, String> properties, String propertyName) {
+		return getPropertyValue(properties, propertyName, null);
+	}
+
+	protected static String getPropertyValue(Map<String, String> properties, String propertyName,
```
cppwfs, line 527: Can we use PropertyParserUtils.getDeploymentPropertyValue instead?

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -85,14 +76,16 @@
...
+	protected DeploymentPropertiesResolver deploymentPropertiesResolver;
+
 	/**
 	 * Create the RuntimeEnvironmentInfo.
 	 *
 	 * @param spiClass the SPI interface class
 	 * @param implementationClass the SPI implementation class
 	 * @return the Kubernetes runtime environment info
 	 */
-	protected RuntimeEnvironmentInfo createRuntimeEnvironmentInfo(Class spiClass, Class implementationClass) {
```
cppwfs, line 95: This method can be left as protected.   Seems like it is only used in child classes.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -125,7 +118,7 @@ protected RuntimeEnvironmentInfo createRuntimeEnvironmentInfo(Class spiClass, Cl
 		return map;
 	}
 
-	protected AppStatus buildAppStatus(String id, PodList podList, ServiceList services) {
+	AppStatus buildAppStatus(String id, PodList podList, ServiceList services) {
```
cppwfs, line 121: This method can be left as protected. Seems like it is only used in child classes.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -156,52 +149,61 @@ protected AppStatus buildAppStatus(String id, PodList podList, ServiceList servi
 		return statusBuilder.build();
 	}
 
+	void logPossibleDownloadResourceMessage(Resource resource) {
```
cppwfs, line 152: This method can be set to protected. Seems like it is only used in child classes.

chrisjs:
no additional comments, nice consolidation but i have seen the following test seems to be failing:

```
[ERROR]   KubernetesAppDeployerIntegrationTests.testDeploymentWithGroupAndIndex:526 
Expected: is "foo"
     but: was null
```

ilayaperumalg:
@chrisjs I got a similar error on my local but I thought this was because of the `LoadBalancer` type being used for the Service. Maybe not?

ilayaperumalg:
> @chrisjs I got a similar error on my local but I thought this was because of the `LoadBalancer` type being used for the Service. Maybe not?

No, it's not. I will get it fixed.

chrisjs:
merged in fbda7b0226ac5e73ae19d87bec0013b1baaac842

PR #382, opened 2020-05-12
Fix Pod container port configuration

### LINKED ISSUE #381: Probes are being applied to tasks
.. and it should not.

ilayaperumalg:
- Set the default `8080` as the container port only for apps deployed via `K8sAppDeployer`
 - In case of task apps, only set the container port if the task configuration has it set
 - Add/update tests

Resolves #381


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -183,9 +182,18 @@ PodSpec createPodSpec(AppDeploymentRequest appDeploymentRequest) {
 
 		ContainerConfiguration containerConfiguration = new ContainerConfiguration(appId, appDeploymentRequest)
 				.withProbeCredentialsSecret(getProbeCredentialsSecret(deploymentProperties))
-				.withExternalPort(externalPort)
 				.withHostNetwork(hostNetwork);
 
+		if (KubernetesAppDeployer.class.isAssignableFrom(this.getClass())) {
+			containerConfiguration.withExternalPort(getExternalPort(appDeploymentRequest));
+		}
+		else {
```
chrisjs, line 190: since before tasks were always setting null for port, is this needed?

ilayaperumalg: Good point. We should only set the port for `K8sAppDeployer` to be in sync with the previous versions.

chrisjs:
merged in db33836b00d85a2db67fe0805f2645513aeb00a6

PR #387, opened 2020-06-15
Update compat chart

### LINKED ISSUE #386: Fix compatibility chart for the latest K8s versions
We need to add the content from https://github.com/spring-io/dataflow.spring.io/pull/256

chrisjs:
Resolves #386

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,13 +3,14 @@ A [Spring Cloud Deployer](https://github.com/spring-cloud/spring-cloud-deployer)
 
 ## Kubernetes Compatibility
 
-| Deployer \ Kubernetes | 1.11 | 1.12 | 1.13 | 1.14 |
-|-----------------------|------|------|------|------|
-| **1.3.x**             | ✓    | ✓    | ✓    | ✕    |
-| **2.0.x**             | ✓    | ✓    | ✓    | ✕    |
-| **2.1.x**             | ✓    | ✓    | ✓    | ✕    |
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
LGTM

sabbyanandan:
@ilayaperumalg: Instead of going back-and-forth between the README and the Microsite with dupes, I'd vote for getting rid of it here completely and cross-link to the Microsite.

ilayaperumalg:
> Instead of going back-and-forth between the README and the Microsite with dupes, I'd vote for getting rid of it here completely and cross-link to the Microsite.

I think it is important to have the matrix in the deployer project as `spring-cloud-deployer-kubernetes` can still be considered a standalone without related to SCDF.

sabbyanandan:
We are never going to keep it up-to-date in both places. If you think it is that important, remove it from the Microsite, and maybe pull the README section on the site.

ilayaperumalg:
> We are never going to keep it up-to-date in both places. If you think it is that important, remove it from the Microsite, and maybe pull the README section on the site.

We could do that at the cost of not showing up the corresponding SCDF version in it. I guess that would not be appropriate. I understand it is difficult to keep it up to date at both the places but this is something we have to do:
 - Show only the deployer compatibility at the Spring Cloud Deployer K8s repo
 - Show SCDF + Deployer K8s compatibility at the Microsite

PR #396, opened 2020-09-15
Add job annotations into cron job builder

### LINKED ISSUE #389: Merging of deployer and scheduler properties at the global level
When setting properties at the server config account level, those properties get populated into `KubernetesDeployerProperties`. This works fine for tasks, streams, etc. When scheduling a task, properties are put into `KubernetesSchedulerProperties` which will work fine with deployment properties, but those specified at the server config account level do not get carried over.

ilayaperumalg:
- Add jobAnnotations by resolving against all the property sources including the Scheduler Properties from the server config
 - Add integration tests to verify the property overriding at the server/deployment levels

Resolves #389


ilayaperumalg:
This needs to be back ported as well.

Issue #401, opened 2020-10-01
Unable to create a pod with "resources: limits" such as "nvidia.com/gpu: 1"


simonbrommers:
At present, the only limits that seem to be able to be applied to created pods are cpu and memory.  Many cases could benefit from being able to specify other types of limits, e.g. nvidia.com/gpu

chrisjs:
functionality is added to the project based on need/requests/submissions/ etc. if you have a particular request please update the issue for the specific part of functionality / use-case. the existing description of "able to specific other types of limits, e.g." is a bit open ended. also, contributions are always welcome so do feel free to make a PR if inclined.

simonbrommers:
Thanks @chrisjs for the suggestion to make my issue more specific.
My project requires the ability to run pods that make use of GPU resources, where we have 1 GPU resource per node (non-sharable).  I have the ability to launch a pod directly within kubectl using the likes of:

`
    spec:
      containers:
      - name: nvidia-pipeline
        image: *********
        imagePullPolicy: IfNotPresent
        resources:
          limits:
           nvidia.com/gpu: 1
`

I require the ability to be able to specify this kind of resource limits when using Spring Cloud Data Flow (either via the .yaml files for deployment or runtime properties).  So far I've looked into making the change, and found method deduceResourceLimits() in class DeploymentPropertiesResolver which is limited to only working with limits.cpu and limits.memory properties, but don't really understand the rest of the framework enough to be able to make the change without the worry of breaking things for others

chrisjs:
thanks for the clarification. that method will look at deployment properties and build a structure which would be used downstream to send that info off to the underlying k8s client lib. similar to the existing cpu/memory, another limit attribute could be added for gpu's in a similar way, possibly structuring that particular resource in a somewhat generic way given nvida and amd gpu's are currently supported with their own name.

simonbrommers:
I was thinking along exactly those lines @chrisjs .  I've got a plan on what code I'd like to change to make this happen.  I'm just struggling at present to be able to build this project and the docker image to test out the changes I want to put in.  Is there a guide to setting up a local development environment in which I can do this change, esp for building the docker image?

chrisjs:
the deployer doesn't need a docker image made to test, that's generally done in the deployer test suite. are you referring to building  dataflow and skipper with the deployer changes and the corresponding image to test end to end?

simonbrommers:
[...]

PR #402, opened 2020-10-05
Fix for issue #401 - Unable to create a pod with "resources: limits" such as "nvidia.com/gpu: 1"

simonbrommers:
My first attempt at a more serious PR for an OS project - please go easy on me..
It's only small in nature - but it fixes a big problem for my use-case - needing to target Spring Cloud Data Flow tasks to GPU capable nodes.
I'm unsure if I'm making this PR to the correct target branch - please advise if not.  My aim is to get the code to a release ASAP.

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
@@ -94,9 +94,11 @@ public class KubernetesDeployerProperties {
 		public LimitsResources() {
 		}
 
-		public LimitsResources(String cpu, String memory) {
+		public LimitsResources(String cpu, String memory, String gpuVendor, String gpuCount) {
 			this.cpu = cpu;
 			this.memory = memory;
+			this.gpuVendor = gpuVendor;
+			this.gpuCount = gpuCount;
 		}
 
 		public String getCpu() {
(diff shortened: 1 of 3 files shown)


simonbrommers:
Thanks for your patience and pointers @chrisjs - I've changed the target branch for the PR to master now, and added tests for the changes I made.  I think the formatting should be ok.. I know I initially made a mess of it, but I think I cleared it up.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -81,12 +81,24 @@ public void setFabric8(Config fabric8) {
 		 */
 		private String memory;
 
+		/**
+		 * Container GPU vendor limit.
```
chrisjs, line 85: "Container GPU vendor name for limit" - this content is extracted and auto populated into UI components, and seems a little more descriptive

chrisjs (changes requested):
this looks good, thanks. i made a couple comments on minor things. once those are patched up ill go ahead and merge it.

simonbrommers:
I've addressed the comments you made now - fingers crossed.
With the hope this will now be accepted and merged, what should I expect in terms of a new release?  Is there something that documents the process that takes place?  I got the feeling milestone builds are in the mix somewhere before an actual release

chrisjs:
squashed and merged as 1e8d60cad804d9afe0531c54bee04e7516ba90ef 

release order is SNAPSHOT -> Milestone(s) -> Release Canadiate(s) -> RELEASE. this change will go into the upcoming RELEASE, @sabbyanandan would be able to provide a better estimate on when that will occur

thanks again for the contrib!

Issue #296, opened 2019-04-24
Limit exposure of fabric8 types in Properties


chrisjs:
Limit exposure of fabric8 types in `KubernetesDeployerProperties` -- ie: `Volume`, `VolumeMount`, along with the `fabric8` property. wrap into our own types / delegate specific properties as needed.

PR #423, opened 2021-03-08
Add support for sidecar container

### LINKED ISSUE #414: Deploying an sidecar container
**Problem description:**
We want to deploy a sidecar container inside the pod where batch-applications running in OpenShift.

**Solution description:**
We want to use a different tool to keep secrets. For this solution we are supposed to deploy a sidecar container.

We are using the 2.7.0-RC1 version of scdf. In this version, we can deploy only one container inside a pod. It is possible to deploy more than one container in a pod?

**Description of alternatives:**
Is it possible to customize default deployment config?

ilayaperumalg:
- Ability to pass sidecar container via PodSpec
 - Applicable whenever PodSpec is created
 - Add tests

Resolves #414

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -258,10 +258,7 @@ public class AbstractKubernetesDeployer {
 		if (initContainer != null) {
 			podSpec.addToInitContainers(initContainer);
 		}
-		List<Container> additionalContainers = this.deploymentPropertiesResolver.getAdditionalContainers(deploymentProperties);
-		for (Container additionalContainer: additionalContainers) {
-			podSpec.addToContainers(additionalContainer);
-		}
+		podSpec.addAllToContainers(this.deploymentPropertiesResolver.getAdditionalContainers(deploymentProperties));
 		return podSpec.build();
 	}
 
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -258,9 +258,9 @@ public class AbstractKubernetesDeployer {
 		if (initContainer != null) {
 			podSpec.addToInitContainers(initContainer);
 		}
-		Container sidecarContainer = this.deploymentPropertiesResolver.getSidecarContainer(deploymentProperties);
-		if (sidecarContainer != null) {
-			podSpec.addToContainers(sidecarContainer);
+		List<Container> additionalContainers = this.deploymentPropertiesResolver.getAdditionalContainers(deploymentProperties);
+		for (Container additionalContainer: additionalContainers) {
+			podSpec.addToContainers(additionalContainer);
 		}
 		return podSpec.build();
 	}
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 the original author or authors.
+ * Copyright 2015-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -258,7 +258,10 @@ public class AbstractKubernetesDeployer {
 		if (initContainer != null) {
 			podSpec.addToInitContainers(initContainer);
 		}
-
+		Container sidecarContainer = this.deploymentPropertiesResolver.getSidecarContainer(deploymentProperties);
+		if (sidecarContainer != null) {
+			podSpec.addToContainers(sidecarContainer);
+		}
 		return podSpec.build();
 	}
 
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
@@ -345,9 +345,8 @@ public class KubernetesDeployerProperties {
 	public static class InitContainer extends ContainerProperties {
 	}
 
-	public static class SidecarContainer extends ContainerProperties {
+	static class Container extends io.fabric8.kubernetes.api.model.Container {
 	}
-
 
 	public static class ContainerProperties {
 		private String imageName;
@@ -697,7 +696,7 @@ public class KubernetesDeployerProperties {
 	/**
 	 * A side car container one can add to the main application container.
 	 */
-	private SidecarContainer sidecarContainer;
+	private List<Container> additionalContainers;
 
 	public String getNamespace() {
 		return namespace;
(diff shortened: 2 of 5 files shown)


isha1kh (approved):

pravinkumarb84 (approved):

chrisjs (changes requested):
i think the terminology here can be more generic. a pod can contain multiple containers. "sidecar" is more of a pattern type term. i think a generic deployer property that contains the container info for one or more containers say something like `additionalContainers` or whatnot would be more appropriate. the property should allow for multiple container definitions to be passed.

ilayaperumalg:
@chrisjs Good point. In fact, when I was getting started with this feature, my original intention was to allow multiple containers instead of just a single additional container. But, since the deployment properties have key/value pairs as String/String, I was thinking maybe we could add the initial support to add a single sidecar container with the specific key named `sidecarContainer` to avoid this key/value constraint. 

Irrespective of this, let's discuss further to find a better approach on this. Thanks!

chrisjs:
> @chrisjs Good point. In fact, when I was getting started with this feature, my original intention was to allow multiple containers instead of just a single additional container. But, since the deployment properties have key/value pairs as String/String, I was thinking maybe we could add the initial support to add a single sidecar container with the specific key named `sidecarContainer` to avoid this key/value constraint.
> 
> Irrespective of this, let's discuss further to find a better approach on this. Thanks!

we already map objects beyond key/value pairs, for example:

https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#_secret_key_references

https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#_configmap_key_references

etc. seems like it would be a similar approach

ilayaperumalg:
ok, yeah, I missed that. I will work on it and update the PR. 

>  i would advocate for using the generic "containers" naming rather than "sidecar"

Sure, `containers` that is :-)

ilayaperumalg:
@chrisjs Updated the PR with the suggested changes. With this set of changes, we can support having **multiple** containers with **all** the supported configurations for a container.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -342,7 +342,13 @@ public Long getFsGroup() {
 		}
 	}
 
-	public static class InitContainer {
+	public static class InitContainer extends ContainerProperties {
+	}
+
+	static class Container extends io.fabric8.kubernetes.api.model.Container {
```
ilayaperumalg, line 348: I am curious if we can use the same approach of using fabric8 api models to other supported properties as well. I believe this can eliminate some explicit models with a fewer properties (for instance the ContainerProperties we have for InitContainer). 

Also, I thought having an explicit type named `Container` could help here. But, I am ok if we can directly use the fabric8 Container type as is in this class.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -258,7 +258,10 @@ PodSpec createPodSpec(AppDeploymentRequest appDeploymentRequest) {
 		if (initContainer != null) {
 			podSpec.addToInitContainers(initContainer);
 		}
-
+		List<Container> additionalContainers = this.deploymentPropertiesResolver.getAdditionalContainers(deploymentProperties);
+		for (Container additionalContainer: additionalContainers) {
+			podSpec.addToContainers(additionalContainer);
```
chrisjs, line 263: looks like theres a `addAllToContainers` that could be used rather than the loop here

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
chrisjs: you could use the fabric8 objects, but thats a bit leakly. theres an open issue around this: https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/296

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -687,6 +693,11 @@ public void setVolumeMounts(List<VolumeMount> volumeMounts) {
 	 */
 	private InitContainer initContainer;
 
+	/**
+	 * A side car container one can add to the main application container.
```
chrisjs, line 697: `A side car container` -> `A container`

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationTests.java
```diff
@@ -196,4 +199,64 @@ public void testDeploymentLabels() {
...
+		String launchId = kubernetesTaskLauncher.launch(request);
+		Timeout timeout = deploymentTimeout();
+
+		assertThat(launchId, eventually(hasStatusThat(
+				Matchers.<TaskStatus>hasProperty("state", Matchers.is(LaunchState.running))), timeout.maxAttempts,
+				timeout.pause));
+
+		String taskName = request.getDefinition().getName();
+
+		log.info("Checking job pod spec annotations of {}...", taskName);
```
chrisjs, line 227: looks like this should be removed, or the message changed

chrisjs (changes requested):
minor things but beyond that LGTM

chrisjs:
merged in d84e6c7071ee9687ee06e363dfe0112d4e9c8abf

ilayaperumalg:
@isha1kh @pravinkumarb84 If possible, it would be great to try out the latest SCDF 2.8.0 SNAPSHOT and Skipper 2.7.0-SNAPSHOT and let us know your feedback on supporting additional containers. Please pay attention to the name of the property `additionalContainers` and you can refer to [this test ](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/423/files#diff-bbb69a3805e1a8fb81588e9fb061b0a78df2e55077d9993ad6426f18fed18397R1213) for how to use it. Thanks!

isha1kh:
@ilayaperumalg - Yes am going to try this soon, thanks.

isha1kh:
@ilayaperumalg - I've tried against the above version and although the additional container is getting added. I've noticed two issues:
1) The additional containers gets into 'CrashLoopBackOff' state and from my analysis (using busybox sample from your test case) it seems that's due to the fact that the container terminates after printing rather than staying active. So its acting more like initContainer than sidecars.

2) When trying to update the stream via Skipper UI, the additional container properties aren't re-populated.

ilayaperumalg:
On the container restart question, You need to set the restart policy for the container spec .

On the UI, you would only see the `additionalContainers` as a root property and won't see the nested properties for the container. Please feel free to work on supporting this feature if you are willing to contribute. Thanks!

umutcann:
Hi @ilayaperumalg, @isha1kh 

I also want to try this implementation. As I understand, we should pass an container yaml as additionalContainers param, right?

Do you have any sample to try?

Thanks.

ilayaperumalg:
You need to set the container properties similar to how we set [here](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/423/files#diff-bbb69a3805e1a8fb81588e9fb061b0a78df2e55077d9993ad6426f18fed18397R1229)

umutcann:
Thanks @ilayaperumalg , I did not notice before :)

PR #427, opened 2021-04-15
Add ability to obtain hostName if IP is not available

### LINKED ISSUE #426: Deployer does not return hostname if external IP is not available
When deploying pods to AWS, AWS does not provide an externalIP but rather  a hostname.    In these cases the deployer needs to return a hostname if the external IP is not available.

This can be observed in the KubernetesAppInstanceStatus:143.

cppwfs:
This is used when obtaining AppInstanceStatus

resolves #426


jvalkeal:
lets try it.

PR #429, opened 2021-04-15
Migrates project to use Fabric8 5.3.0 [Do Not Merge]

### LINKED ISSUE #428: Update Deployer to use Fabric8 5.3.0
Before merging make sure that Skipper and Dataflow are also queued up for this change.

cppwfs:
resolves #428

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
 
 	<properties>
 		<java.version>1.8</java.version>
-		<kubernetes-client.version>4.10.2</kubernetes-client.version>
+		<kubernetes-client.version>5.3.0</kubernetes-client.version>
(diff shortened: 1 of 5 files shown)


chrisjs:
the original intent of this upgrade was to test out token expiration fixes that were made in later versions. while we noticed some improvement, the issue still remains. 

given its not fully fixed and there is an upcoming release, we may consider first trying auth with an SA token instead of exec'ing to tmc. if that works, we should open an issue with fabric8 in hopes we can get a proper fix in, test and then upgrade preferably at the start of a release cycle

cppwfs:
I will close this PR and we can handle the upgrade later.

PR #432, opened 2021-04-30
Enable Http probe scheme configuration

tzolov:
- Allow configuring the livenessProbe.scheme with livenessHttpProbeScheme and readinessProbe.scheme with he readinessHttpProbeScheme properties.
 - Scheme defaults to HTTP. But can be set to HTTPS for SSL applications.


ilayaperumalg:
LGTM, merging.

PR #447, opened 2021-05-21
Add VolumeClaimTemplate name override support

### LINKED ISSUE #418: Support custom name of volumeClaimTemplates for statefulSet deployments
While we have deployment properties to specify size and class of the volumeClaimTemplate for a stateful set (`statefulSet.volumeClaimTemplate.storageClassName` and `statefulSet.volumeClaimTemplate.storage`), there doesn't seem to be a way to specify the template's name. Instead it is hardcoded to be the appId.

Now, in order to really make use of the persistent storage, we would need to add a volumeMount referencing the template by name. When deploying via Spring Cloud Dataflow, I have not been able to figure out a way to predictably infer the `appId` upfront. Am I missing something?

I would like to suggest something like `statefulSet.volumeClaimTemplate.name` so that we have a way to deterministically reference it in a volumeMount.

ilayaperumalg:
Resolves #418

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
@@ -274,10 +274,14 @@ public class KubernetesAppDeployer extends AbstractKubernetesDeployer implements
 
 		String storageClassName = this.deploymentPropertiesResolver.getStatefulSetStorageClassName(kubernetesDeployerProperties);
 
+		String volumeClaimTemplateName = this.deploymentPropertiesResolver.getStatefulSetVolumeClaimTemplateName(kubernetesDeployerProperties);
+
+		volumeClaimTemplateName = StringUtils.hasText(volumeClaimTemplateName) ? volumeClaimTemplateName : appId;
+
 		PersistentVolumeClaimBuilder persistentVolumeClaimBuilder = new PersistentVolumeClaimBuilder().withNewSpec().
 				withStorageClassName(storageClassName).withAccessModes(Collections.singletonList("ReadWriteOnce"))
 				.withNewResources().addToLimits(storageResource).addToRequests(storageResource).endResources()
-				.endSpec().withNewMetadata().withName(appId).withLabels(idMap)
+				.endSpec().withNewMetadata().withName(volumeClaimTemplateName).withLabels(idMap)
 				.addToLabels(SPRING_MARKER_KEY, SPRING_MARKER_VALUE).endMetadata();
 
 		PodSpec podSpec = createPodSpec(request);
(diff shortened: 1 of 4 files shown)


src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployerIntegrationIT.java
```diff
@@ -925,6 +926,47 @@ public void createStatefulSetWithOverridingRequest() throws Exception {
...
+		log.info("Deploying {}...", appDeploymentRequest.getDefinition().getName());
+		String deploymentId = deployer.deploy(appDeploymentRequest);
+		Timeout timeout = deploymentTimeout();
+		await().pollInterval(Duration.ofMillis(timeout.pause))
+				.atMost(Duration.ofMillis(timeout.maxAttempts * timeout.pause))
+				.untilAsserted(() -> {
+					assertThat(appDeployer().status(deploymentId).getState()).isEqualTo(DeploymentState.deployed);
+				});
+
+		Map<String, String> selector = Collections.singletonMap(SPRING_APP_KEY, deploymentId);
```
tzolov, line 949: I get a Compilation failure of missing `SPRING_APP_KEY`

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
```diff
@@ -275,10 +275,14 @@ protected void createStatefulSet(AppDeploymentRequest request) {
 
 		String storageClassName = this.deploymentPropertiesResolver.getStatefulSetStorageClassName(kubernetesDeployerProperties);
 
+		String volumeClaimTemplateName = this.deploymentPropertiesResolver.getStatefulSetVolumeClaimTemplateName(kubernetesDeployerProperties);
+
+		volumeClaimTemplateName = StringUtils.hasText(volumeClaimTemplateName) ? volumeClaimTemplateName : appId;
```
tzolov, line 280: If not mistaken the `appId`s are used in by the app runtime information returned to the clients and perhaps by skipper. 
Is there a risk for the above change to affect any existing skipper behavior?

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployerIntegrationIT.java
ilayaperumalg: Rebase would fix this. Will update it.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployer.java
ilayaperumalg: The `appId` is still the default and it's importance as the name of the volume  here is just being unique for the app being deployed. Otherwise, we provide the option to make the volume name configurable.

tzolov: got it

ilayaperumalg:
Updated the PR after rebase

PR #457, opened 2021-07-14
Deprecate scheduler properties

cppwfs:
Deprecate the use of scheduler properties.   All scheduling properties should come from deployment properties.
Child of https://github.com/spring-cloud/spring-cloud-dataflow/issues/4262
Resolves SCDK-456

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
@@ -465,6 +465,59 @@ public class KubernetesDeployerProperties {
 	}
 
 	/**
+	 * The {@link RestartPolicy} to use. Defaults to {@link RestartPolicy#Never}.
+	 */
+	private RestartPolicy restartPolicy = RestartPolicy.Never;
+
+	/**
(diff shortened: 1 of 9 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -464,6 +464,59 @@ public void setEnvironmentVariables(String[] environmentVariables) {
...
+	/**
+	 * The default service account name to use for tasks.
+	 */
+	protected static final String DEFAULT_TASK_SERVICE_ACCOUNT_NAME = "default";
+
+	/**
+	 * Service account name to use for tasks, defaults to:
+	 * {@link #DEFAULT_TASK_SERVICE_ACCOUNT_NAME}
+	 */
+	private String taskServiceAccountName = DEFAULT_TASK_SERVICE_ACCOUNT_NAME;
```
ilayaperumalg, line 481: I believe the changes from this PR can fix this as well: https://github.com/spring-cloud/spring-cloud-dataflow/issues/4619

ilayaperumalg:
LGTM. We need to spend more time manually testing the various use-cases. Merging this to get started with the testing.

PR #458, opened 2021-09-09
Add Supplemental Groups to Pod Security Context.

umutcann:
Supplemental Groups which controls group IDs containers will be a new feature for pod security context.


dturanski:
@umutcann - Thank you for the contribution. In reviewing it, I noticed a bug in `DeployerPropertiesResolver.getPodSecurityContext` that was there originally. 
I am closing this and replacing with #459 which is cleaner.  Thanks again!

PR #461, opened 2021-09-23
Add Supplemental Groups to Pod Security Context.

### LINKED ISSUE #459: Supplemental Groups Need for Pod Security Context
Hi,

We are running file processing batch application on distributed Kubernetes environments. In order to [enable](https://stackoverflow.com/a/67868742/14420141) restartability on Kubernetes environments for batch apps which include file operations, we are using persistent volume with NFS.

But we are facing permission error during access different clusters. According to our consultant, we need to set [SupplementalGroups](https://docs.openshift.com/container-platform/4.8/storage/persistent_storage/persistent-storage-nfs.html#storage-persistent-storage-nfs-group-ids_persistent-storage-nfs) value over SCDF. But SCDF is not supporting Supplemental Groups value for now.

I opened a pull-request to support that.

https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/458

May you please overview?

Thanks.

dturanski:
Fixes #459 Rebase, refactor and merge #458 Also fixes bug in DeploymentPropertiesResolver


ilayaperumalg:
LGTM, merging.

PR #464, opened 2021-11-03
Task logs with multiple containers

### LINKED ISSUE #463: Error getting task logs for multiple containers
We have run Spring Batch in GKE through SCDF and we use a sidecar container in this batch, so we have two container in pods. When I launch a task a try to get logs I get error in SCDF:

"a container name must be specified for pod task-3o8kogmxmg, choose one of"

I think that SCDF should get logs from both container.

jvalkeal:
- Add same way to get logs as its done on a stream side.
- Fixes #463


cppwfs:
LGTM
Kicked the tires using multiple containters.

PR #471, opened 2021-12-02
Deployer properties are no longer filtered for scheduling

### LINKED ISSUE #467: Service account is not getting set
@gurugubelli commented on [Tue Jul 06 2021](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4619)

Even though environment variable SPRING_CLOUD_SCHEDULER_KUBERNETES_TASK_SERVICE_ACCOUNT_NAME exists used to work in spring cloud dataflow 2.5.x but  stopped working in 2.8.1. Service account getting set default always rather than custom account.

---

@ilayaperumalg commented on [Mon Jul 12 2021](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4619#issuecomment-878212570)

Hi @gurugubelli, Thanks for reporting this. This is a bug and we'll try to address this in the upcoming release. Since this needs to be fixed at the SCDF level, I need to move this there. 

---

@ilayaperumalg commented on [Mon Jul 12 2021](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4619#issuecomment-878223461)

You can work around this problem by passing the scheduler properties when creating the scheduler for the task in SCDF as follows:

```
scheduler.kubernetes.taskServiceAccountName=<sa-name>
```

---

@ilayaperumalg commented on [Wed Nov 10 2021](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4619#issuecomment-964948772)

@cppwfs Could you confirm if the changes from https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/457 fixes this issue please?

---

@cppwfs commented on [Wed Nov 10 2021](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4619#issuecomment-965553705)

Hello @ilayaperumalg This appears to be another issue.    We will need to investigate further.   

---

@ilayaperumalg commented on [Fri Nov 12 2021](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4619#issuecomment-967101180)

Yes, after some investigation I found that the [DeploymentPropertiesResolver](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/main/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DeploymentPropertiesResolver.java) in the K8s deployer still has the references ([here](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/e2d29194bc871e70fdad7cde001f6e4a4159dbc5/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DeploymentPropertiesResolver.java#L628) and [here](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/e2d29194bc871e70fdad7cde001f6e4a4159dbc5/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DeploymentPropertiesResolver.java#L632)) to the deprecated K8s scheduler properties which need to be fixed. 

Also,[ from the scheduler properties deprecation efforts](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/457), the K8s scheduler switched to use the K8s deployer properties instead. Hence, the resolution for the issue reported in here is eventually fixing the support for `taskServiceAccountName` as a K8s deployer property set as platform properties instead of scheduler property.

---

@cppwfs commented on [Fri Nov 12 2021](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4619#issuecomment-967114354)

@ilayaperumalg thanks for looking into this.   I have also assigned myself to this issue.

cppwfs:
resolves #467

### DIFF
--- a/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/DeploymentPropertiesResolverTests.java
+++ b/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/DeploymentPropertiesResolverTests.java
@@ -37,7 +37,7 @@ public class DeploymentPropertiesResolverTests {
 		DeploymentPropertiesResolver deploymentPropertiesResolver = getDeploymentPropertiesResolver(isDeprecated, kubernetesDeployerProperties);
 		Map<String, String> properties = new HashMap<>();
 		RestartPolicy restartPolicy = deploymentPropertiesResolver.getRestartPolicy(properties);
-		Assertions.assertThat(restartPolicy).isEqualTo(RestartPolicy.Always);
+		Assertions.assertThat(restartPolicy).isEqualTo(RestartPolicy.Never);
 		if (isDeprecated) {
 			properties.put(KubernetesSchedulerProperties.KUBERNETES_SCHEDULER_PROPERTIES_PREFIX + ".restartPolicy", RestartPolicy.Never.name());
 		}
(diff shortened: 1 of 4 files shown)


src/test/java/org/springframework/cloud/deployer/spi/kubernetes/DeploymentPropertiesResolverTests.java
```diff
@@ -37,7 +37,7 @@ public void testRestartPolicy(boolean isDeprecated) {
 		DeploymentPropertiesResolver deploymentPropertiesResolver = getDeploymentPropertiesResolver(isDeprecated, kubernetesDeployerProperties);
 		Map<String, String> properties = new HashMap<>();
 		RestartPolicy restartPolicy = deploymentPropertiesResolver.getRestartPolicy(properties);
-		Assertions.assertThat(restartPolicy).isEqualTo(RestartPolicy.Always);
+		Assertions.assertThat(restartPolicy).isEqualTo(RestartPolicy.Never);
```
ilayaperumalg, line 40: Currently, with the scheduler properties deprecation we have got the RestartPolicy set to `Never` by [default](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/main/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java#L563). @cppwfs @onobc Could you double check if this doesn't have any impact with our Stream apps from the SCDF side?

onobc (approved):

Issue #412, opened 2021-01-21
seccompProfile under podSecurityContext does not work


ilayaperumalg:
@TimoSchaz commented on [Thu Jan 21 2021](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4340)

I tried to apply seccompProfile under podSecurityContext for Tasks and Streams that should be deployed via SCDF Server and Skipper ,but it seems that they are not used.
Is it possible to set the Profile there ?
I think this is an important feature that should be supportet to provide better security.

I configured it as follows but its not working:

data:
  application.yaml: |-
    spring:
      cloud:
        skipper:
          server:
            platform:
              kubernetes:
                accounts:
                  default:
                    podSecurityContext:
                      seccompProfile:
                            type: Localhost
                            localhostProfile: profile.json

---

@ilayaperumalg commented on [Thu Jan 21 2021](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4340#issuecomment-764556710)

Hi @TimoSchaz , Thanks for reporting!

Yes, `seccompProfile` is not supported and we can look at supporting in the upcoming release.
If you would like to contribute, please submit a PR and we'd happy to help you clarify any questions you may have.

I am also moving this to `spring-cloud-deployer-kubernetes` as this issue belongs there.

TimoSchaz:
Thank you for the fast feedback.
Another thing that also seems to be not implemented yet is the support for securityContext
as follows

 securityContext:
            allowPrivilegeEscalation: false
            readOnlyRootFilesystem: true

To be able to make the started tasks and streams filesystems readonly this is needed.

Is this possible at the moment?
If not i think allowPrivilegeEscalation: false and readOnlyRootFilesystem: true should be implemented, because they are important security features.

ilayaperumalg:
@TimoSchaz Sure, please go ahead and submit a PR to accommodate these properties. We'd be happy to help if you have any questions.

PR #472, opened 2021-12-13
Add support for seccompProfile on pod security context and configuring container security context.

bono007:
@ilayaperumalg 
~The motivating issue also [calls out](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/412#issuecomment-769017511) the need for container level `securityContext.(allowPrivilegeEscalation|readOnlyRootFilesystem)` support . However, I would prefer if we split that into another ticket (or at least an additional pull request) to keep the changes in this code proposal concise~

| ℹ️ Initially I was going to break this into 2 separate pull requests but decided to combine them once I saw them both implemented. |
| --- | 

### Adds support for configuring the security context on the main container

- **Question 1:** I only apply the context to the main container. Should we add this to the initContainer as well? There is no need to do anything special for additional containers as the only mechanism to configure those is via full yaml - consumers can simply add the security context in that yaml if need be. 
- **Question 2:** Should we support more than these 2 requested properties? I know we do not support all available properties on the pod security context either. On one hand, adding more properties increases code to maintain. On the other hand, adding more properties gives users more options. 

### Refactors KubernetesAppDeployerTests as follows:
- Moved the pod and container security related tests into their own `@Nested` tests. 
- Simplified the assertions to use actual/expected object equality rather than individual property asserts
- I implemented the new container security context tests using the `@Nested` approach and it looked like a great fit for the existing pod security context tests as well. Because re-ordering and renaming code during a review makes for a not so fun review, I intentionally made small commits (2-6) to show step by step what I did w/ the pod security context tests to get to the final result. I apologize in advance for any grief this causes the reviewer. These of course should all be squashed into a single commit into main.

**Motivation:** 
The `KubernetesAppDeployerTests` has grown quite large and it is a challenge for a newbie (me) to grok all of whats going on in there. I leveraged `@Nested` to group the pod and container security context tests. I really love the output that `@Nested` produces in IntelliJ as well (see below)

nested-test-output

- **Question 3:** I am curious to what the team thinks on this test as a whole. Should it be:
  - left as-is (don't use `@Nested`)?
  - split into separate `KubernetesAppDeployer<tested-area-goes-here>Tests`?
  - apply @Nessted to other areas of the test as well? 

### Remaining items 
- [ ] Finish updating the counterpart pull request to [spring-cloud-dataflow-docs](https://github.com/spring-cloud/spring-cloud-dataflow/pull/4784).

### DIFF
--- a/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployerTests.java
+++ b/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployerTests.java
@@ -26,6 +26,7 @@ import java.util.Properties;
 
 import io.fabric8.kubernetes.api.model.AffinityBuilder;
 import io.fabric8.kubernetes.api.model.ConfigMapKeySelector;
+import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.EnvVar;
 import io.fabric8.kubernetes.api.model.HostPathVolumeSource;
 import io.fabric8.kubernetes.api.model.HostPathVolumeSourceBuilder;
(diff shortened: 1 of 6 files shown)


ilayaperumalg:
@bono007 Thank you for the contribution! Sure, we can have those other properties addressed via a separate ticket.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployerTests.java
```diff
@@ -1149,95 +1094,6 @@ public void testPodAntiAffinityFromYaml() throws Exception {
...
-		AppDeploymentRequest appDeploymentRequest = new AppDeploymentRequest(definition, getResource(), props);
-
-		deployer = new KubernetesAppDeployer(new KubernetesDeployerProperties(), null);
-		PodSpec podSpec = deployer.createPodSpec(appDeploymentRequest);
-
-		PodSecurityContext podSecurityContext = podSpec.getSecurityContext();
-
-		assertThat(podSecurityContext).as("Pod security context should not be null").isNotNull();
-
-		assertThat(podSecurityContext.getRunAsUser()).as("Unexpected run as user").isNull();
```
bono007, line 1187: I also remove the `as` from the assertions as I have found over time these become a maintenance pain and are not much more helpful over assertJ default message and line numbers. This is the approach taken in Spring Boot. I am not sure if this is a hard rule in Spring Cloud (this repo in particular) or not.

dturanski:
LGTM; Merging.

PR #473, opened 2021-12-18
Throw exception if pod or job is not exist

umutcann:


cppwfs:
Hello @umutcann,
Thanks for the contribution!    I tested the solution and it appears that the exception never is thrown.   This is because `podsToDelete` is not null and the `podsToDelete.list().getItems()` returns an empty list.    Keep in mind the test I ran had other pods running in the cluster.

bono007:
Hi @umutcann ,

> ... This is because `podsToDelete` is not null and the `podsToDelete.list().getItems()` returns an empty list. 

I ❤️ [ObjectUtils.isEmpty](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/ObjectUtils.html#isEmpty-java.lang.Object-) as it checks for both null and empty. We could adjust the `if` to 
```java
if (jobsToDelete != null && !ObjectUtils.isEmpty(jobsToDelete.list().getItems()) {
```

Hi @cppwfs ,

This would probably be a good test case to add. However, I don't think we have any places that **directly** test the cleanup. The abstract IT cleans up after each test and does **exercise** the cleanup. Would it be worth adding a test specifically for cleanup verification? I'm not sure if this case warrants it or not.

cppwfs:
Hey @bono007!
I agree that is a great test to add.

bono007:
> Hey @bono007! I agree that is a great test to add.

@cppwfs 
Sounds good. I will submit a PR and ping back here when I do.

bono007:
Hi @umutcann 


Hi

I changed the code locally to use `ObjectUtils` and verified the before/after w/ the following test added to `org.springframework.cloud.deployer.spi.kubernetes.KubernetesTaskLauncherIntegrationIT`

```java
@Test
void cleanupForNonExistentTaskThrowsException() {
  KubernetesTaskLauncher kubernetesTaskLauncher = new KubernetesTaskLauncher(new KubernetesDeployerProperties(),
      new KubernetesTaskLauncherProperties(), kubernetesClient);
  assertThatThrownBy(() -> kubernetesTaskLauncher.cleanup("foo"))
      .isInstanceOf(IllegalStateException.class)
      .hasMessage("Cannot delete Pod for task \"%s\" (reason: pod does not exist)", "foo");
}
```

Do you mind adding this to your code proposal? 

Thanks

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
```diff
@@ -393,6 +393,9 @@ private void deleteJob(String id) {
 			boolean jobDeleted = jobsToDelete.delete();
 			logger.debug(String.format("Job deleted for: %s - %b", id, jobDeleted));
 		}
+		else {
```
bono007, line 396: I think by doing the empty check first we can simplify this a bit (ditch the else) such as:
```java
if (podsToDelete == null || ObjectUtils.isEmpty(podsToDelete.list().getItems())) {
  throw new IllegalStateException(String.format("Cannot delete Pod for task \"%s\" (reason: pod does not exist)", id));
}

logger.debug(String.format("Deleting Pod for task: %s", id));
boolean podsDeleted = podsToDelete.delete();
logger.debug(String.format("Pod deleted for: %s - %b", id, podsDeleted));
```

Wdyt?

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
```diff
@@ -404,6 +407,9 @@ private void deletePod(String id) {
 			boolean podsDeleted = podsToDelete.delete();
 			logger.debug(String.format("Pod deleted for: %s - %b", id, podsDeleted));
 		}
+		else {
+			throw new IllegalStateException(String.format("Cannot delete Pod for task: %s. Because Pod is not exist!", id));
```
bono007, line 411: I think we can replace the "!" with a reason phrase at the end of the message wrapped in parens to make it stand out. Something like:
```java
throw new IllegalStateException(String.format("Cannot delete Pod for task \"%s\" (reason: pod does not exist)", id));
```

The output before:
```
Cannot delete Pod for task: foo. Because Pod is not exist!
```
The output after: 
```
Cannot delete Pod for task "foo" (reason: pod does not exist)
```
Wdyt @umutcann ?

bono007:
> > Hey @bono007! I agree that is a great test to add.
> 
> @cppwfs Sounds good. I will submit a PR and ping back here when I do.

Hi @cppwfs - I added the IT [here](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/475). Please take a scan when you can. Thx.

onobc:
Closed via the https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/commit/8349b8c503b151884a810349569210c6b3e673d8 (the PR commit had the wrong issue in it so this did not get auto-resolved).

PR #475, opened 2021-12-26
Add integration tests for task cleanup (jobs and pods)

bono007:
Adds an integration test for task cleanup for both job and pod based tasks. 

Motivated by https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/473#issuecomment-1000327922

My original intent was to add a single test for the cleanup method. However, once I got into the code I saw an opportunity for refactoring a bit (explained inline in the review self-comments). I apologize in advance for the "extra" changes I have introduced. My intention was to layer it step/step in commits to illustrate the evolution more clearly. However, at some point I just kept coding and looked up and said "oops, I forgot to slow-step commit along the way".

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,9 @@ gcloud container --project {your-project-name} clusters create "spring-test" --z
 gcloud config set container/cluster spring-test
 gcloud config set compute/zone us-central1-b
 gcloud container clusters get-credentials spring-test
+kubectl version
 ```
+> :information_source: the last command causes the access token to be generated and saved to the kubeconfig file - it can be any valid kubectl command
 
 #### Running the tests
 
--- a/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesTaskLauncherIntegrationTests.java
+++ b/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesTaskLauncherIntegrationTests.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2021-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.deployer.spi.kubernetes;
+
+import java.lang.reflect.Method;
+import java.time.Duration;
+import java.util.List;
+import java.util.UUID;
+
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.batch.v1.Job;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import org.awaitility.core.ConditionFactory;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.cloud.deployer.resource.docker.DockerResource;
+import org.springframework.cloud.deployer.spi.task.TaskLauncher;
+import org.springframework.cloud.deployer.spi.test.AbstractTaskLauncherIntegrationJUnit5Tests;
+import org.springframework.cloud.deployer.spi.test.Timeout;
+import org.springframework.core.io.Resource;
+
+import static org.awaitility.Awaitility.await;
+
+/**
+ * Abstract base class for integration tests for {@link KubernetesTaskLauncher}.
+ *
+ * @author Chris Bono
+ */
+abstract class AbstractKubernetesTaskLauncherIntegrationTests extends AbstractTaskLauncherIntegrationJUnit5Tests {
+
+	@Autowired
+	@SuppressWarnings("SpringJavaInjectionPointsAutowiringInspection")
+	protected TaskLauncher taskLauncher;
+
+	@Autowired
+	@SuppressWarnings("SpringJavaInjectionPointsAutowiringInspection")
+	protected KubernetesClient kubernetesClient;
+
+	@Override
+	protected TaskLauncher provideTaskLauncher() {
+		return taskLauncher;
+	}
+
+	@Override
+	protected String randomName() {
+		return "task-" + UUID.randomUUID().toString().substring(0, 18);
+	}
+
+	@Override
+	protected Resource testApplication() {
+		return new DockerResource("springcloud/spring-cloud-deployer-spi-test-app:latest");
+	}
+
+	@Override
+	protected Timeout deploymentTimeout() {
+		return new Timeout(20, 5000);
+	}
+
+	@Test
+	@Override
+	@Disabled("Currently reported as failed instead of cancelled")
+	public void testSimpleCancel() throws InterruptedException {
+		super.testSimpleCancel();
+	}
+
+	protected void logTestInfo(TestInfo testInfo) {
+		log.info("Testing {}...", testInfo.getTestMethod().map(Method::getName).orElse(testInfo.getDisplayName()));
+	}
+
+	protected ConditionFactory awaitWithPollAndTimeout(Timeout timeout) {
+		return await().pollInterval(Duration.ofMillis(timeout.pause))
+				.atMost(Duration.ofMillis(timeout.maxAttempts * timeout.pause));
+	}
+
+	protected List<Pod> getPodsForTask(String taskName) {
+		return kubernetesClient.pods().withLabel("task-name", taskName).list().getItems();
+	}
+
+	protected List<Job> getJobsForTask(String taskName) {
+		return kubernetesClient.batch().v1().jobs().withLabel("task-name", taskName).list().getItems();
+	}
+}
--- a/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationIT.java
+++ b/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationIT.java
@@ -40,6 +40,9 @@ import static org.assertj.core.api.Assertions.entry;
 /**
  * Integration tests for {@link KubernetesTaskLauncher}.
  *
+ * <p>NOTE: The tests do not call {@code TaskLauncher.destroy/cleanup} in a finally block but instead rely on the
+ * {@link AbstractKubernetesTaskLauncherIntegrationTests#cleanupLingeringApps() AfterEach method} to clean any stray apps.
+ *
  * @author Thomas Risberg
  * @author Chris Schaefer
  * @author Chris Bono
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -40,7 +40,9 @@ gcloud container --project {your-project-name} clusters create "spring-test" --z
 gcloud config set container/cluster spring-test
 gcloud config set compute/zone us-central1-b
 gcloud container clusters get-credentials spring-test
+kubectl version
 ```
+> :information_source: the last command causes the access token to be generated and saved to the kubeconfig file - it can be any valid kubectl command
```
bono007, line 45: So this was a "fun" one... I was banging my head wondering why the auth would work sometime and not others. We have to actually run `kubectl` with a command that requires auth in order for the auth token to be written to the kubeconfig file. To see the proof, do the following:

```java
gcloud container clusters get-credentials spring-test
kubectl config view
```

> ℹ️  note that `kubectl config view` does not suffice to get the token generated but `kubectl version` does.

You will notice there is no "access-token" entry in the "users" section for the "spring-test" cluster - the end of the kubeconfig file will look similar to:
```yaml
- name: gke_bono007-sandbox_us-central1-b_spring-test
  user:
    auth-provider:
      config:
        cmd-args: config config-helper --format=json
        cmd-path: /Users/cbono/indeed/spring.io/scdf/google-cloud-sdk/bin/gcloud
        expiry-key: '{.credential.token_expiry}'
        token-key: '{.credential.access_token}'
      name: gcp
```
and your tests will fail auth.

Now execute the following commands:
```java
kubectl version
kubectl config view
```
You will notice there is now an "access-token" entry in the "users" section for the "spring-test" cluster - the end of the kubeconfig file will now look similar to:
```yaml
- name: gke_bono007-sandbox_us-central1-b_spring-test
  user:
    auth-provider:
      config:
        access-token: <redacted>
        cmd-args: config config-helper --format=json
        cmd-path: /Users/cbono/indeed/spring.io/scdf/google-cloud-sdk/bin/gcloud
        expiry: "2021-12-26T04:55:03Z"
        expiry-key: '{.credential.token_expiry}'
        token-key: '{.credential.access_token}'
      name: gcp
```
and your tests will now pass.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesTaskLauncherIntegrationTests.java
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.core.io.Resource;
+
+import static org.awaitility.Awaitility.await;
+
+/**
+ * Abstract base class for integration tests for {@link KubernetesTaskLauncher}.
+ *
+ * @author Chris Bono
+ */
+abstract class AbstractKubernetesTaskLauncherIntegrationTests extends AbstractTaskLauncherIntegrationJUnit5Tests {
```
bono007, line 46: I ended up pulling the common bits out of the pod and job flavors of the ITs. I could go either way on this as it does help w/ DRY but it is not a ton of fields/methods and it would not be the end of the world to have each concrete IT have this same content repeated in each.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesTaskLauncherIntegrationTests.java
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	@Test
+	@Override
+	@Disabled("Currently reported as failed instead of cancelled")
+	public void testSimpleCancel() throws InterruptedException {
+		super.testSimpleCancel();
+	}
+
+	protected void logTestInfo(TestInfo testInfo) {
```
bono007, line 83: This is here to allow the specific tests to not have to maintain the string that it was printing out at the top of the method such as 
```java
log.info("Testing {}...", "<insert-hardcoded-test-name-here>");
```
It leverages the JUnit `TestInfo` that is available for injection on every test method.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationIT.java
```diff
@@ -16,235 +16,122 @@
...
-	protected Resource testApplication() {
-		return new DockerResource("springcloud/spring-cloud-deployer-spi-test-app:latest");
-	}
-
-	@Override
-	protected Timeout deploymentTimeout() {
-		return new Timeout(20, 5000);
+	void taskLaunchedWithJobPodAnnotations(TestInfo testInfo) {
+		logTestInfo(testInfo);
+		launchTaskPodAndValidateCreatedPodWithCleanup(
```
bono007, line 55: One thing I noticed about both the job and pod ITs is that they all did the following:
1. setup an app def/request 
2. launch the task
3. wait for the task to be running
4. verify the condition(s) under test
5. destroy the task 

I also noticed only steps 1 and 4 changed on each test. And for "step 1" it was only a different set of deployment props. 

To pull those steps out and only write them once I introduced this `launchTaskPodAndValidateCreatedPodWithCleanup` that takes in set of props (uses those when doing step 1 above) and then takes in a consumer of the resulting pod which is responsible for the assertions on the pod. 

I think this makes the test more concise and hopefully easier to add tests in the future.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationIT.java
```diff
@@ -16,235 +16,122 @@
...
+					assertThat(pod.getMetadata().getAnnotations()).isNotEmpty()
+							.contains(entry("key1", "val1"), entry("key2", "val2"), entry("key3", "val31:val32"));
+				});
 	}
 
 	@Test
-	public void testJobPodAnnotation() {
-		log.info("Testing {}...", "JobPodAnnotation");
-
-		KubernetesTaskLauncher kubernetesTaskLauncher = new KubernetesTaskLauncher(new KubernetesDeployerProperties(),
```
bono007, line 95: Also notice that we were constructing this local KTL in order to pass in the KDP that we modified in the job-based IT but do not do anything to in the pod-based IT. This is really not needed since this is a `@SpringBootTest` that is auto-configured w/ `KubernetesAutoConfiguration` which results in a KTL that is plenty sufficient to use. In the job-based IT, a `@SpringBootTest( properties = "` is what varies the KTL between job/pod based tasks. 

This also allows us to get rid of some extra code.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationIT.java
```diff
@@ -16,235 +16,122 @@
...
 	}
 
 	@Test
-	public void testTaskAdditionalContainer() {
-		log.info("Testing {}...", "TaskAdditionalContainer");
-
-		KubernetesTaskLauncher kubernetesTaskLauncher = new KubernetesTaskLauncher(new KubernetesDeployerProperties(),
-				new KubernetesTaskLauncherProperties(), kubernetesClient);
-
+	void cleanupDeletesTaskPod(TestInfo testInfo) {
```
bono007, line 113: And this one test was the whole motivation for this code proposal. It varies little w/ the main flow of the others besides that fact that it calls `cleanup` directly, rather than `destroy`. It then, validates that it can't find the pods after the delete - which seems to be the best we can do to verify that the pod was actually deleted. 

And once I got to this point I realized we could have easily mocked the PodList/JobList code out and done this w/ a simple unit test. All we really want to know is that "when I call cleanup(taskId) that the underlying task job/podList has `delete` called on it as well. Of course the downside of that approach is the mocking required which drives us pretty deep down into a specific impl of api calls to get back the pod/job list etc.. I am inclined to leave this verification here in the IT based on this fact of the unit test mock approach being more brittle. However, I am open to switching to unit test path if others feel strongly in favor of the unit test approach.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesTaskLauncherIntegrationTests.java
```diff
@@ -0,0 +1,99 @@
...
+		return await().pollInterval(Duration.ofMillis(timeout.pause))
+				.atMost(Duration.ofMillis(timeout.maxAttempts * timeout.pause));
+	}
+
+	protected List<Pod> getPodsForTask(String taskName) {
+		return kubernetesClient.pods().withLabel("task-name", taskName).list().getItems();
+	}
+
+	protected List<Job> getJobsForTask(String taskName) {
+		return kubernetesClient.batch().v1().jobs().withLabel("task-name", taskName).list().getItems();
```
bono007, line 97: I also slipped this in here so wanted to point it out.. 

I replaced the deprecated `batch().jobs()` w/ the contemporary `batch().v1().jobs()`. If this is an issue for some reason I am unaware of just lmk.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationIT.java
```diff
@@ -16,235 +16,122 @@
...
-		log.info("Checking job pod spec labels of {}...", taskName);
-
-		List<Pod> pods = kubernetesClient.pods().withLabel("task-name", taskName).list().getItems();
+		log.info("Launching {}...", taskName);
+		String launchId = taskLauncher().launch(request);
+		awaitWithPollAndTimeout(deploymentTimeout())
+				.untilAsserted(() -> assertThat(taskLauncher().status(launchId).getState()).isEqualTo(LaunchState.running));
 
+		log.info("Checking task Pod for {}...", taskName);
+		List<Pod> pods = getPodsForTask(taskName); //kubernetesClient.pods().withLabel("task-name", taskName).list().getItems();
```
cppwfs, line 102: Probably remove this comment.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationIT.java
```diff
@@ -16,235 +16,122 @@
...
-
-		List<String> commands = testAdditionalContainer.getCommand();
-
-		assertThat(commands).contains("sh", "-c", "echo hello");
-
-		log.info("Destroying {}...", taskName);
-
-		timeout = undeploymentTimeout();
-		kubernetesTaskLauncher.destroy(taskName);
+		log.info("Cleaning up {}...", taskName);
```
cppwfs, line 129: What are your thoughts on having this on a ` finally` block ?   So we can attempt a cleanup if something in the test fails.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationIT.java
bono007: Good eyes! Thank you.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherIntegrationIT.java
bono007: 👋🏻 @cppwfs ,

I definitely considered this case and should have left a comment. Here is why I left it w/o the finally:

```java 
TLW = TaskLauncherWrapper
```

This test is similiar to all of the others w/ the exception that it calls `TLW.cleanup()` rather than `TLW.destroy()`. The other tests rely on the parent's `@AfterEach` which handles any stragglers:
```java
@AfterEach
public void cleanupLingeringApps() {
  for (String id : launcherWrapper.launchedTasks) {
    try {
      launcherWrapper.wrapped.cleanup(id);
    }
    catch (Exception e) {
      log.warn("Exception caught while trying to cleanup '{}'. Moving on...", id);
    }
  }
  for (String appName : launcherWrapper.deployedApps) {
    try {
      log.warn("Test named '{}' left behind an app for ''. Trying to destroy.", this.testName, appName);
      launcherWrapper.wrapped.destroy(appName);
    }
    catch (Exception e) {
      log.warn("Exception caught while trying to destroy '{}'. Moving on...", appName);
    }
  }
}
```
I believe this will handle this new test case as well.

cppwfs: Doh!   Thanks for the explanation,  I forgot to check the parent!

cppwfs:
LGTM

cppwfs:
Squashed Merged

cppwfs:
Thanks for the contribution!

PR #482, opened 2022-02-09
Updated so that deployer can handle args with out assignment operator

### LINKED ISSUE #481: Task fails to launch if argument does not have assignment operator
If a user specifies a single value argument (without `=` with subsequent value) the deployer throws an exception.
Deployer should be able to parse args with out an assignment value.
For example this should be parsible:
`a=b c d=e`

Stack Trace is available here: 
https://github.com/spring-cloud/spring-cloud-dataflow/issues/4805

cppwfs:
resolves #481

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2021 the original author or authors.
+ * Copyright 2015-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -274,15 +275,9 @@ public class DefaultContainerFactory implements ContainerFactory {
 	List<String> createCommandArgs(AppDeploymentRequest request) {
 		List<String> cmdArgs = new LinkedList<>();
 
-		List<String> commandArgOptions = new ArrayList<>();
-		for(String arg : request.getCommandlineArguments()) {
-			int indexOfAssignment = arg.indexOf("=");
-			if(indexOfAssignment < 0) {
-				commandArgOptions.add(arg);
-			} else {
-				commandArgOptions.add(arg.substring(0, indexOfAssignment).replaceAll("^--",""));
-			}
-		}
+		List<String> commandArgOptions = request.getCommandlineArguments().stream()
+		.map(this::getArgOption)
+		.collect(Collectors.toList());
 
 		// add properties from deployment request
 		Map<String, String> args = request.getDefinition().getProperties();
@@ -307,6 +302,11 @@ public class DefaultContainerFactory implements ContainerFactory {
 		return cmdArgs;
 	}
 
+	private String getArgOption(String arg) {
+		int indexOfAssignment = arg.indexOf("=");
+		String argOption = (indexOfAssignment < 0) ? arg : arg.substring(0, indexOfAssignment);
+		return argOption.trim().replaceAll("^--", "");
+	 }
 
 	private DeploymentPropertiesResolver getDeploymentPropertiesResolver(AppDeploymentRequest request) {
 		String propertiesPrefix = (request instanceof ScheduleRequest &&
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
@@ -24,7 +24,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
```diff
@@ -54,6 +53,7 @@
  * @author David Turanski
  * @author Chris Schaefer
  * @author Ilayaperumal Gopinathan
+ * @author Glenn Renfro
```
onobc, line 56: Do you mind bumping the copyright as well?

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactoryTests.java
```diff
@@ -1421,6 +1421,22 @@ public void testCommandLineArgsOverridesExistingProperties() {
 		assertThat(defaultContainerFactory.createCommandArgs(appDeploymentRequest)).containsExactly("--foo=newValue");
 	}
 
+	@Test
+	public void testCommandLineArgsNoAssignment() {
+		List<String> args = new ArrayList();
+		args.add("a");
+		args.add("b = c");
```
onobc, line 1428: Good test to add. 

Should we also beef up the "override from app definition" testcase to verify that the `b = c` case is handled properly? Its checking on app keys and now we have spaces etc.. (ie "b " != "b")

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
```diff
@@ -273,9 +273,16 @@ private ProbeType getProbeType(ContainerConfiguration containerConfiguration) {
 	 */
 	List<String> createCommandArgs(AppDeploymentRequest request) {
 		List<String> cmdArgs = new LinkedList<>();
-		List<String> commandArgOptions = request.getCommandlineArguments().stream()
-				.map(arg-> arg.substring(0,arg.indexOf("=")).replaceAll("^--",""))
```
onobc, line 277: **[OPINION-NITatom]** I think leaving this as a simple stream+map but have the map use a method reference that hides all of the "how to get the arg" would be easier to follow. 

Something like: 
```java
List<String> commandArgOptions = request.getCommandlineArguments().stream()
    .map(this::getArgOption)
    .collect(Collectors.toList());
}

private String getArgOption(String arg) {
   int indexOfAssignment = arg.indexOf("=");
   String argOption = (indexOfAssignment < 0) ? arg : arg.substring(0, indexOfAssignment);
   return argOption.trim().replaceAll("^--", "");
}
```

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
```diff
@@ -273,9 +273,16 @@ private ProbeType getProbeType(ContainerConfiguration containerConfiguration) {
...
-				.collect(Collectors.toList());
+
+		List<String> commandArgOptions = new ArrayList<>();
+		for(String arg : request.getCommandlineArguments()) {
+			int indexOfAssignment = arg.indexOf("=");
+			if(indexOfAssignment < 0) {
+				commandArgOptions.add(arg);
```
onobc, line 281: Should we also be scrubbing this option-only-variant for `--`?

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
```diff
@@ -273,9 +273,16 @@ private ProbeType getProbeType(ContainerConfiguration containerConfiguration) {
...
+			} else {
+				commandArgOptions.add(arg.substring(0, indexOfAssignment).replaceAll("^--",""));
+			}
+		}
```
onobc, line 285: Should we also scrub the arg for extra whitespace (trim) so that it matches up against overrides from app definition?

onobc (commented):
@cppwfs looks good. I have a few minor comments.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
cppwfs: Sure.  What year is it? :-D

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactoryTests.java
cppwfs: Sounds good.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
cppwfs: Its only stripped to make sure we are comparing like properties.   If I understand the question.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
cppwfs: For now I'd say no.   If there is a need we can certainly strip it later.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
onobc: I meant should we also make sure we strip the `--` from the `key-only` style args like we are doing w/ the `key=value` style args. The latest commit handles both and your tests cover all of the cases I can think of so I will resolve this now.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
onobc: The current commit is actually doing the `trim()` as it was in my suggested code for that method. Just making sure you are ok w/ that.

onobc (approved):

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DefaultContainerFactory.java
cppwfs: Ahh.  gotcha.   
I went back and forth.  I guess I said yes by using the code you suggested.

onobc:
Closed via https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/commit/370a0a140ce53affd1cb9f78f94a48775fdf5bf5

Issue #484, opened 2022-02-15
Error occurs if use arguments without '='


LOG-INFO:
If I use an argument without '=', an error occurs.

ex) `test1 test2=test2`
expected arguement list `["test1", "test2=test2"]`, but just an error occurs.

related issues: https://github.com/spring-cloud/spring-cloud-dataflow/issues/4828#issuecomment-1039387258

#### Exception StackTrace
> ```
> java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 4
>         at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3319)
>         at java.base/java.lang.String.substring(String.java:1874)
>         at org.springframework.cloud.deployer.spi.kubernetes.DefaultContainerFactory.lambda$createCommandArgs$0(DefaultContainerFactory.java:277)
>         at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
>         at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
>         at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
>         at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
>         at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
>         at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
>         at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
>         at org.springframework.cloud.deployer.spi.kubernetes.DefaultContainerFactory.createCommandArgs(DefaultContainerFactory.java:278)
>         at org.springframework.cloud.deployer.spi.kubernetes.DefaultContainerFactory.create(DefaultContainerFactory.java:102)
>         at org.springframework.cloud.deployer.spi.kubernetes.AbstractKubernetesDeployer.createPodSpec(AbstractKubernetesDeployer.java:210)
>         at org.springframework.cloud.deployer.spi.kubernetes.KubernetesScheduler.createCronJob(KubernetesScheduler.java:176)
>         at org.springframework.cloud.deployer.spi.kubernetes.KubernetesScheduler.schedule(KubernetesScheduler.java:75)
[...]

PR #485, opened 2022-02-15
fix(#484): make it possible to use non key-value argument

LOG-INFO:


onobc:
Hi @LOG-INFO 
Do you mind adding a test (or two) that covers this change? Also, please rebase as we have merged a couple of other PRs underneath this one. 

Thanks

onobc:
Thanks @LOG-INFO for the follow up and closing it out.

PR #502, opened 2022-07-15
Add supporing of ttlSecondsAfterFinished configuration

alexpodolyak:
This pull request is a proposal to add a supporting of automatic clean-up for finished jobs: https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/
 
*Changelog*:
 - Add supporting of ```ttlSecondsAfterFinished``` job spec. configuration
 - Test coverage

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherProperties.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherProperties.java
@@ -36,6 +36,13 @@ public class KubernetesTaskLauncherProperties {
 	private Integer backoffLimit;
 
 	/**
+	 * The number of seconds after a job has finished before it is eligible to be automatically
+	 * removed by the TTL controller - note that logs from removed jobs will not be able to
+	 * be retrieved.
+	 */
+	private Integer ttlSecondsAfterFinished;
+
+	/**
 	 * Obtains the {@link RestartPolicy} to use. Defaults to
 	 * {@link KubernetesTaskLauncherProperties#restartPolicy}.
 	 *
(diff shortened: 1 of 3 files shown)


onobc (changes requested):
Thanks for the contribution @alexpodolyak . Code looks great and thanks for adding the IT  - nice work!

The only comment I have is around updating the comment. After that we should be able to merge this.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncherProperties.java
```diff
@@ -35,6 +35,11 @@ public class KubernetesTaskLauncherProperties {
 	 */
 	private Integer backoffLimit;
 
+	/**
+	 * The time to life value to limit the lifetime of Jobs that have finished execution.
```
onobc, line 39: Do you mind updating this slightly to include the caveat that removed jobs will not be able to have their logs retrieved. Something like this:

```
	 * The number of seconds after a job has finished before it is eligible to be automatically removed by the TTL 
	 * controller - note that logs from removed jobs will not be able to be retrieved.
```

onobc: Also, you have a checkstyle failure    https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/runs/7437762079?check_suite_focus=true#step:6:990

alexpodolyak: Thanks for Your comments @onobc. I'd change the comment and fix the checkstyle issue.

onobc (approved):
Looks good @alexpodolyak . Thanks again for the contribution.

PR #530, opened 2023-02-22
Add ttlSecondsAfterFinished cronjob spec to KubernetesScheduler

angelzard83:
Added ttlSecondsAfterFinished setting to automatically delete jobs 
when creating cronjob in spring-cloud-dataflow.

### DIFF
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesScheduler.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesScheduler.java
@@ -216,10 +216,12 @@ public class KubernetesScheduler extends AbstractKubernetesDeployer implements S
 			concurrencyPolicy = "Allow";
 		}
 
-		Integer ttlSecondsAfterFinished = null;
+		final Integer ttlSecondsAfterFinished;
 		String ttlSecondsAfterFinishedString = schedulerProperties.get(KUBERNETES_DEPLOYER_CRON_TTL_SECONDS_AFTER_FINISHED);
-		if (StringUtils.hasText(ttlSecondsAfterFinishedString) && Pattern.matches("^[0-9]+$", ttlSecondsAfterFinishedString)) {
+		if (StringUtils.hasText(ttlSecondsAfterFinishedString)) {
 			ttlSecondsAfterFinished = Integer.parseInt(ttlSecondsAfterFinishedString);
+		} else {
+			ttlSecondsAfterFinished = this.properties.getCron().getTtlSecondsAfterFinished();
 		}
 
 		PodSpec podSpec = createPodSpec(new ScheduleRequest(scheduleRequest.getDefinition(),schedulerProperties, scheduleRequest.getCommandlineArguments(), scheduleRequest.getScheduleName(),scheduleRequest.getResource()));
--- a/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesScheduler.java
+++ b/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesScheduler.java
@@ -52,8 +52,10 @@ public class KubernetesScheduler extends AbstractKubernetesDeployer implements S
 	protected static final String SPRING_CRONJOB_ID_KEY = "spring-cronjob-id";
 
 	private static final String SCHEDULE_EXPRESSION_FIELD_NAME = "spec.schedule";
-	
+
 	static final String KUBERNETES_DEPLOYER_CRON_CONCURRENCY_POLICY = KubernetesDeployerProperties.KUBERNETES_DEPLOYER_PROPERTIES_PREFIX+".cron.concurrencyPolicy";
+
+	static final String KUBERNETES_DEPLOYER_CRON_TTL_SECONDS_AFTER_FINISHED = KubernetesDeployerProperties.KUBERNETES_DEPLOYER_PROPERTIES_PREFIX+".cron.ttlSecondsAfterFinished";
 
 	public KubernetesScheduler(KubernetesClient client,
 			KubernetesDeployerProperties properties) {
@@ -65,8 +67,8 @@ public class KubernetesScheduler extends AbstractKubernetesDeployer implements S
 		this.containerFactory = new DefaultContainerFactory(properties);
 		this.deploymentPropertiesResolver = new DeploymentPropertiesResolver(
 				(properties instanceof KubernetesSchedulerProperties) ?
-				KubernetesSchedulerProperties.KUBERNETES_SCHEDULER_PROPERTIES_PREFIX:
-				KubernetesDeployerProperties.KUBERNETES_DEPLOYER_PROPERTIES_PREFIX, properties);
+						KubernetesSchedulerProperties.KUBERNETES_SCHEDULER_PROPERTIES_PREFIX:
+						KubernetesDeployerProperties.KUBERNETES_DEPLOYER_PROPERTIES_PREFIX, properties);
 	}
 
 	@Override
@@ -204,7 +206,7 @@ public class KubernetesScheduler extends AbstractKubernetesDeployer implements S
 				schedulerProperties.get("spring.cloud.deployer.kubernetes.cron.expression") :
 				schedulerProperties.get(SchedulerPropertyKeys.CRON_EXPRESSION);
 		Assert.hasText(schedule, "The property spring.cloud.deployer.cron.expression must be defined");
-		
+
 		String concurrencyPolicy = schedulerProperties.get(KUBERNETES_DEPLOYER_CRON_CONCURRENCY_POLICY);
 		// check default server properties
 		if (!StringUtils.hasText(concurrencyPolicy)) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesScheduler.java
```diff
@@ -204,15 +206,21 @@ protected CronJob createCronJob(ScheduleRequest scheduleRequest) {
...
 		if(concurrencyPolicy==null) {
 			concurrencyPolicy = "Allow";
-		} 
+		}
+
+		Integer ttlSecondsAfterFinished = null;
+		String ttlSecondsAfterFinishedString = schedulerProperties.get(KUBERNETES_DEPLOYER_CRON_TTL_SECONDS_AFTER_FINISHED);
+		if (StringUtils.hasText(ttlSecondsAfterFinishedString) && Pattern.matches("^[0-9]+$", ttlSecondsAfterFinishedString)) {
+			ttlSecondsAfterFinished = Integer.parseInt(ttlSecondsAfterFinishedString);
+		}
```
onobc, line 223: We typically check the deployer props then drop down to the default server props when not set (see [concurrencyPolicy](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/530/files#diff-901dcb4f8a1f2bf100d52f408bc9e357f0d9b0af8375930b7713f8c62ffabadbR211) above). 

Please add the `ttlSecondsAfterFinished` property to the deployer properties `CronConfig` object and then default to that, something like: 
```java
Integer ttlSecondsAfterFinished;
String ttlSecondsAfterFinishedStr = schedulerProperties.get(KUBERNETES_DEPLOYER_CRON_TTL_SECONDS_AFTER_FINISHED);
if (StringUtils.hasText(ttlSecondsAfterFinishedStr)) {
	ttlSecondsAfterFinished = Integer.parseInt(ttlSecondsAfterFinishedStr);
}
else {
	ttlSecondsAfterFinished = this.properties.getCron().getTtlSecondsAfterFinishedString();
}
```

You will also see that I removed the pattern validation. The reason is that It is better to let it fail during parse which will tell the user "You have an invalid property setting". Otherwise it will just skip the parse and the user will wonder why there ttl setting is not taking effect. Eventually, sometime later, after pulling their hair out, they will see that they accidentally set the value to something like `"10'"` when it was supposed to be `"10"`.  😄

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesScheduler.java
```diff
@@ -226,7 +234,8 @@ protected CronJob createCronJob(ScheduleRequest scheduleRequest) {
 		CronJob cronJob = new CronJobBuilder().withNewMetadata().withName(scheduleRequest.getScheduleName())
 				.withLabels(labels).withAnnotations(this.deploymentPropertiesResolver.getJobAnnotations(schedulerProperties)).endMetadata()
 				.withNewSpec().withSchedule(schedule).withConcurrencyPolicy(concurrencyPolicy).withNewJobTemplate()
-				.withNewSpec().withNewTemplate().withNewMetadata().addToAnnotations(annotations).addToLabels(labels)
+				.withNewSpec().withTtlSecondsAfterFinished(ttlSecondsAfterFinished)
```
onobc, line 237: We will need to come up w/ a good default in the (yet to be added) `this.properties.getCron().getTtlSecondsAfterFinishedString();` or not set the ttl value on the builder when its null as I believe the underlying builder will not like a null value here.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesSchedulerIT.java
```diff
@@ -965,7 +965,7 @@ public void testTaskServiceAccountNameDefault(boolean isDeprecated) {
 	
 	@ParameterizedTest
 	@ValueSource(strings = {"Forbid","Allow","Replace"})
-	public void testConcurencyPolicy(String concurrencyPolicy) {
```
onobc, line 968: Nice catch. Thanks.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesSchedulerIT.java
```diff
@@ -988,6 +988,34 @@ public void testConcurencyPolicy(String concurrencyPolicy) {
 		kubernetesScheduler.unschedule(cronJob.getMetadata().getName());
 	}
 
+	@ParameterizedTest
+	@ValueSource(strings = {"100", "86400"})
```
onobc, line 992: Nice test. I can think of a couple of other cases we could add to improve coverage:
* invalid ttl value (deployer prop)
* invalid ttl value (server default prop)
* deployer prop set and server default prop set (deployer prop wins)
* no deployer prop set but server default prop set
* no deployer prop set and no server default prop set
Wdyt?

onobc (changes requested):
Thanks @angelzard83 for the contribution - nice work. There are a couple of minor comments to address. Once those are resolved I think this one is good to go.

angelzard83:
@onobc 
Thank you for your review.
I added the ttlSecondsAfterFinished property to the deployer properties CronConfig object and removed the pattern check and more tests as you reviewed.
I think that ttlSecondsAfterFinished property should be null as default property because it is optional property.
Please give me more review to this committed code.

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesSchedulerIT.java
onobc: Well done.

src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesScheduler.java
```diff
@@ -204,15 +206,23 @@ protected CronJob createCronJob(ScheduleRequest scheduleRequest) {
...
 			concurrencyPolicy = "Allow";
-		} 
+		}
+
+		final Integer ttlSecondsAfterFinished;
+		String ttlSecondsAfterFinishedString = schedulerProperties.get(KUBERNETES_DEPLOYER_CRON_TTL_SECONDS_AFTER_FINISHED);
+		if (StringUtils.hasText(ttlSecondsAfterFinishedString)) {
+			ttlSecondsAfterFinished = Integer.parseInt(ttlSecondsAfterFinishedString);
+		} else {
+			ttlSecondsAfterFinished = this.properties.getCron().getTtlSecondsAfterFinished();
```
onobc, line 224: Did you forget to add a file? ;) 

https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/actions/runs/4264049330/jobs/7422743930#step:6:3054

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesSchedulerIT.java
```diff
@@ -195,12 +195,12 @@ public void testSchedulerPropertiesMerge() {
 
 		assertThat(mergedProperties
 				.get(KubernetesDeployerProperties.KUBERNETES_DEPLOYER_PROPERTIES_PREFIX + ".imagePullPolicy"))
-						.as("Expected value from Scheduler properties, but found in Deployer properties")
-						.isEqualTo("Never");
+				.as("Expected value from Scheduler properties, but found in Deployer properties")
```
onobc, line 198: Typically we try to not do widespread formatting changes in a PR. 
If it must be done, then it is best to do it in a separate commit to help decipher the actual change from the noise. 
In this case, thank you for updating the formatting here as it really needed it and your changes are small enough that it is not an issue. Just a request for the future - thats all.

onobc (changes requested):
@angelzard83 thx for the updates - well done. I will mark as "Request changes" only because we need to missing file added (see comments). Otherwise this one is approved and well done. 

I do have 1 follow on ask if you have time and don't mind..

Do you mind updating the Dataflow ref docs https://github.com/spring-cloud/spring-cloud-dataflow/blob/main/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-kubernetes-app-properties.adoc to add the new property? I know that having the docs in this repo would be more discoverable but they live out-of-band for now. 

Thanks again

src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesSchedulerIT.java
angelzard83: I'm sorry about that. I'll remember and and practice it.

onobc: Oh no worries - nothing to apologize for. Someone passed that on to me at some point and I am passing it onto you. You can then pass it on to someone as well. 👍🏻

onobc (approved):

onobc:
Nice work @angelzard83 ! 
Thanks again for the contribution and also for the follow up doc PR in dataflow - much appreciated.

onobc:
Closing via https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/commit/d76a971a839a8c05208e0acac771d94e9c33458e

